PR #30, opened 2020-11-17
[PD-6476] - location url and location input format

jeremytubbs:
- Adds new `/locations/:location/jobs` endpoint - http://localhost:8080/locations/flower-mound-tx/jobs
- Adds a helper to format the location input from a slug


src/components/Search/AppSearchProvider.vue
```diff
@@ -172,11 +173,23 @@ export default {
...
+
         setInputFromQuery() {
             this.input = merge(this.getInputDefaults(), this.$route.query)
-
             this.formatInput()
         },
+
+        setInputFromParams (query) {
+            let params = { ... this.$route.params }
+            Object.entries(params).map(([key, value]) => {
```
aljordan82, line 187: I don't think Object.entries is supported in IE11, 
are you using a polyfill?

jeremytubbs: According to https://caniuse.com/object-entries a polyfill is included in core-js which is in our dependencies https://github.com/zloirock/core-js#ecmascript-object

aljordan82: There shouldn't be a need to import it if corejs is part of the build.

jeremytubbs: Yeah, I believe it is included via the `@vue/cli-plugin-babel/preset` defined in `babel.config.js`

src/components/Search/AppSearchProvider.vue
```diff
@@ -172,11 +173,23 @@ export default {
...
+            this.input.location = this.input.location ? displayLocationFromSlug(this.input.location) : ""
+        },
+
         setInputFromQuery() {
             this.input = merge(this.getInputDefaults(), this.$route.query)
-
             this.formatInput()
         },
+
+        setInputFromParams (query) {
```
surgiie, line 185: `query` is not being used anywhere in the function

src/services/helpers.js
```diff
@@ -69,6 +69,12 @@ export function buildJobDetailUrl(title, location, guid) {
...
 const slugify = (string) => (
     words(toString(string).replace(/["\u2019+:]/g, ""), /[\w]+/g).reduce((result, word, index) => (
-      result + (index ? "-" : "") + word.toLowerCase()
+        result + (index ? "-" : "") + word.toLowerCase()
     ), "")
-  )
\ No newline at end of file
+)
+
+export function displayLocationFromSlug(string) {
```
surgiie, line 76: Im assuming at some point we will be undoing slugs for more than just location so i think we can make a more generic function for this maybe one that looks unslugifies (is that a word?) and accepts a delimiter for in between words?

ex `unslugify(location, ",")`

src/components/Search/AppSearchProvider.vue
jeremytubbs: removed

surgiie (approved):

src/services/helpers.js
jeremytubbs: No this is also capitalizes the last... and hopefully I will not need to do this anywhere else.

PR #34, opened 2020-12-02
[PD-6900] - old SEO urls

jeremytubbs:
These urls do not work... they just can exist for now. 

Actually http://localhost:8080/Sales%20Representative/jobs-in but it is an exact match which is the same behavior as http://localhost:8080/job-titles/Sales%20Representative/jobs. Need to work on further url slug formatting in another PR.

- only allows filters to be three levels deep
- builds the urls in the same order as old SEO urls
- adds and old url constant to be used in building the url combinations

### DIFF
--- a/src/services/gridsomeOldSeoPaths.js
+++ b/src/services/gridsomeOldSeoPaths.js
@@ -1,0 +1,30 @@
+const _ = require("lodash")
+const config = require("./../config.js")
+const oldSeoPaths = require("./../constants/oldSeoPaths.js")
+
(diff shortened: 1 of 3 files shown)


src/services/gridsomeOldSeoPaths.js
```diff
@@ -0,0 +1,30 @@
+const _ = require("lodash")
```
surgiie, line 1: Is there a reason you are importing all of lodash instead of just union? 


```
import { union } from 'lodash'

or
import union from 'lodash/get'
```
Not sure how either affects build size or performance.

surgiie (approved):

PR #36, opened 2020-12-03
Cleanups to search filter + other minor refactors.

surgiie:
* Rename accordion props  `isOpen` and `active` to be more consistent with other component naming.
* Decouple Accordion and Search Filter components - This allows implementer to decide if filters should be in accordions or not.
* Remove all formatting code from filter component, most of the data is now in the format we want but remaining formatting has been moved to back end:
https://github.com/DirectEmployers/microsites-search-api/pull/49


jeremytubbs (approved):
I see some issues with autocomplete and would like to see some different formatting of the moc titles. All of which can be addressed if we/when we add moc_ac to solr.

PR #37, opened 2020-12-08
[PD-6913] - Break search provider into separate components

nate998877:
### What does this PR do?
Breaks AppSearchProvider into two separate files AppGoogleTalentSearchProvider and AppSolrSearchProvider

- Replicated functionality of AppSearchProvider Including Solr fallback
- Updated config to use google_talent API (May want to reject/revert)
- Refactored relevant components to use updated providers (App404, Index, jobs)
- change SumbitSearchForm to newSearch and refactored all relevant components as part of the provider breakup now uses emit events to trigger
- removed AppJobDescription from AppJob
- added default value to getSavedTalentSearch in events.js
- refactored SearchService class to use function and removed async await syntax
- refactored CommuteSearchService class to use function and removed async await syntax

## Notes to Reviewer
While minor updates were done in various relevant files. The review should focus on ensuring all functionality was properly ported to the new provider components.

## Future Considerations
- Pull in Vuex
- Break up job wrapper component into separate components
- Refactor services into plain javascript for consistency
- Remove unused components
- Remove location services (Note: possibly unused)


src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -0,0 +1,123 @@
...
+                filters.push({
+                    display: `Commute:${commuteLocation}`,
+                    parameter: "commuteLocation",
+                })
+            }
+            return filters
+        }
+    },
+    methods: {
+        getCommuteDefaults(){
```
surgiie, line 68: Looking at this now in more detail, i think you can just remove this function's return to `inputDefaults`?

src/components/AppChip.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <span class="chip" tabindex="0" @click="clickedChip">
+    <span class="chip" tabindex="0" @click="clickedChip" v-once>
```
surgiie, line 2: Would it better to add the `v-once` directive on the component directly instead of in the component? 

This gives the implementer more control over whether the component should be rendered only once?

jeremytubbs (changes requested):
Issue with double section tags being rendered for `AppGoogleTalentSearchProvider` 
image
Also renders two nested section tags if I replace with `AppSolrSearchProvider`.

It would be preferred if we used the `<components :is="">` in jobs.vue for example to determine which provider to render and try to change imports to be lazy loaded.  This would make it so we were not nesting solr inside of google_talent when switching the config source. 

```
components: {
    AppGoogleTalentSearchProvider: () => import('./AppGoogleTalentSearchProvider.vue'),
    AppSolrSearchProvider: () => import('./AppSolrSearchProvider.vue')
  }
```

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -0,0 +1,89 @@
...
+    </component>
+</template>
+<script>
+import {
+    searchService,
+    commuteSearchService,
+} from "../../../services/search"
+import base from "./mixins/provider"
+import {blank} from "../../../services/helpers"
+import AppSolrSearchProvider from "./AppSolrSearchProvider"
```
surgiie, line 30: Unused import.

src/demo/components/App404.vue
```diff
@@ -3,22 +3,23 @@
     <div class="bg-gray-100 mt-6 p-2">
         <div class="container px-4 mx-auto mt-6 font-semibold">
             <div class="mt-6 search-form">
-              <AppSearchProvider
-                   :site-config="$siteConfig"
+                <AppGoogleTalentSearchProvider
```
surgiie, line 6: Looks like we forgot to update reference here to use `<component>`

surgiie (changes requested):

jeremytubbs (changes requested):
Solr fallback / Google talent config issues:
- When I break google talent by changing the "project_id" to an invalid value the front end makes two successful requests to solr for jobs for the fallback data.
- When I break google talent by changing the `tenant_uuid` to be invalid for some reason the search result is still  returns a `200` response with no jobs. 
- When I try to break google_talent by providing no `company_uuids` and providing only and empty array I still get 200 response with jobs from google talent.

All of the above should result in a single request for fallback solr jobs.

nate998877:
> Solr fallback / Google talent config issues:
> 
> * When I break google talent by changing the `project_id` to an invalid value the front end makes two successful requests to solr for jobs for the fallback data.
> * When I break google talent by changing the `tenant_uuid` to be invalid for some reason the search result is still  returns a `200` response with no jobs.
> * When I try to break google_talent by providing no `company_uuids` and providing only and empty array I still get 200 response with jobs from google talent.
> 
> All of the above should result in a single request for fallback solr jobs.

* Fixed issue where invalid Project_ID causes two post requests
* Invalid Tenant_uuid issues caused by backend issue. Will resolve in future PR
* Providing no company_uuid and still getting jobs is an intended feature of the API as you might want to return all of a given tenant's jobs without having to specify all of the company's uuids

src/components/Search/Providers/mixins/provider.js
```diff
@@ -0,0 +1,218 @@
...
+export default  {
+    props:{
+        siteConfig: {
+            required: true,
+            type: Object,
+        },
+        tag: {
+            required: false,
+            type: String,
+            default: "section",
```
jeremytubbs, line 14: change this to "div"

src/components/Search/AppSearchFilter.vue
```diff
@@ -12,22 +12,23 @@
...
         >
             <ul class="search-filter-options">
                 <li
-                    class="search-filter-options-item"
                     :key="index"
+                    class="search-filter-options-item"
                     v-for="(option, index) in displayedFilters"
                 >
-                    <g-link :to="option.href">
+                    <span @click="selectOption(option)">
```
jeremytubbs, line 28: These should still be links.

src/pages/jobs.vue
```diff
@@ -143,16 +134,16 @@
 
                                         <ul class="search-filter-options">
                                             <li
-                                                class="search-filter-options-item"
                                                 :key="index"
+                                                class="search-filter-options-item"
                                                 v-for="(filter, index) in displayedFilters"
                                             >
-                                                <g-link :to="filter.href">
+                                                <span @click="selectFilter(filter)">
```
jeremytubbs, line 141: This should also remain a link.

src/pages/jobs.vue
```diff
@@ -96,33 +84,35 @@
...
                                 </template>
                                 <ul>
                                     <li
-                                        @click="sort(option)"
+                                        @click="sort.sortField(option)"
                                         class="cursor-pointer"
                                         :key="index"
-                                        v-for="(option, index) in sortOptions"
+                                        v-for="(option, index) in sort.options"
                                         name="sort"
```
jeremytubbs, line 103: `name` is meaningless on an li?

jeremytubbs (approved):

PR #39, opened 2020-12-16
Minor fixes for rendering errors and search filter component not filtering out commute input on regular searches.

surgiie:
Fixes several rendering issues and commute input not cleared on filter urls.

See comments.


src/components/Search/AppSearchFilter.vue
```diff
@@ -104,21 +104,24 @@ export default {
...
             displayedFilters: {},
         }
     },
     created() {
-
+        let value = null
         let filteredOptions = []
         this.givenOptions.forEach(option => {
-            if(this.input[this.name] != option.display){
+            value = this.optionHasSubmitValue(option) ? option.submit: option.display
```
surgiie, line 115: This code just allows options to have a "submit" property on them to allow to specify what values should be submitting. We were already doing this before but i think i accidently removed it in some recent refactors. This allows the component to be used manually or for explicit filters not coming back from back end. Ex:

```html
     <AppSearchFilter
            name="r"
            display="Radius"
            :input="filteredInput"
            :options="[
                { display: 'Exact Location Only', submit: ''},
                { display: 'Within 25 Miles', submit: 25 },
                { display: 'Within 50 Miles', submit: 50 }
            ]"
        >
        </AppSearchFilter>
```

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -17,6 +17,7 @@
             :appliedFilters="appliedFilters"
             :isCommuteSearch="isCommuteSearch"
             :getFilterOptions="getFilterOptions"
+            :filteredInput="getCurrentPayload()"
```
surgiie, line 20: Returns relevant data/input based on the search type being completed.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -160,7 +161,7 @@ export default  {
         },
         getFilterOptions(filter) {
             let key = blank(filter.key) ? filter.name : filter.key
-            return this.filters[key] || []
+            return (blank(this.filters[key]) || !Array.isArray(this.filters[key])) ? [] : this.filters[key]
```
surgiie, line 164: For some reason when you dont have mini kube running or you get a timeout `this.filters[key] || []` evaluates to empty sometimes other times Vue was returning a `mutator` function and that caused rendering issues with the prop validation on AppSearchFilter expecting an array for `options`

src/components/Search/Providers/mixins/provider.js
```diff
@@ -60,7 +60,8 @@ export default  {
             return this.siteConfig.filters || []
         },
         sort() {
-            let sort = {}
+            let sort = { options: [], sortField: ()=>{}, by: null}
```
surgiie, line 63: Define some defaults so the html doesnt error out when no response comes back for sort meta.

src/components/Search/AppSearchFilter.vue
jeremytubbs, line 115: Can you add a space before the `:` of your ternary.

src/components/Search/AppSearchFilter.vue
```diff
@@ -104,21 +104,24 @@ export default {
...
+            if(this.name == 'r'){
+                console.log(this.input, value)
```
jeremytubbs, line 117: console.log?

src/components/Search/Providers/mixins/provider.js
jeremytubbs: Please format this better with new lines

jeremytubbs (approved):

PR #41, opened 2020-12-16
[PD-6886] load more scroll pagination

nate998877:
## What does this PR do?
adds a button that loads more content on the page

## Notes to Reviewer
A few bugs are known to exist, but either require future PRs or can be resolved with VueX.
- When using index.vue to do a search when load more is first clicked add params are added to URL. This doesn't appear to happen anyway else. This can be fixed easily with vuex.
- when initially loading a page with ?page=# all jobs from page 1 - page # are loaded leading to timeouts. This is a backend bug.


jeremytubbs (changes requested):
Changes:
- rename `AppButtonPagination` to `AppLoadMorePagination`
- remove `pagination_type` from config.js
- do not use if-else `showButton` just render import and use preferred type of pagination.
- add a prop to provider.js `isLoadMore` with default of false to be used instead of the siteConfig
- move button above "# jobs found"
- add some style to the button

Improvement:
- prefetch next page of jobs to allow quicker load more experience.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -101,6 +109,26 @@ export default  {
         applyFilters(){
             return []
         },
+        loadMore(page) {
+            //This code is a potential solution for timeouts. A backend PR will be required to implement
+            // let getOffset = (items_per_page) => {
```
jeremytubbs, line 114: Go ahead and create a backend PR to solve this.

jeremytubbs:
Closing due to changes in requirements... my bad!

PR #43, opened 2020-12-17
Provider Improvements/Improve filter clearing.

surgiie:
Made some comments on cleanups, but let me know if you are against the changes or have questions.


src/components/AppPagination.vue
```diff
@@ -151,7 +151,7 @@ export default {
 
             this.current = page
 
-            this.$emit("pageSelected", page)
+            this.$emit("pageSelected", { page: page })
```
surgiie, line 154: See commented about `setInput` below.

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -1,91 +1,76 @@
...
-            :selectPage="selectPage"
-            :featuredJobs="featuredJobs"
-            :removeFilter="removeFilter"
-            :isGoogleTalent="isGoogleTalent"
-            :appliedFilters="appliedFilters"
-            :isCommuteSearch="isCommuteSearch"
-            :getFilterOptions="getFilterOptions"
-            :filteredInput="getCurrentPayload()"
-        ></slot>
+        <slot v-bind="slotData"></slot>
```
surgiie, line 3: Helps not having to sync slot data in each provider.

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -1,91 +1,76 @@
...
         },
     },
     methods: {
-        queryChanged(){
-            this.isCommuteSearch = this.checkIsCommuteSearch()
+        queryChanged() {
+            this.isCommuteSearch = this.shouldDoCommuteSearch()
         },
-        inputDefaults() {
+        providerInputDefinition() {
```
surgiie, line 26: There are several inputs that are coupled together in regards to filter chip removal that introduced specific if logic in the `removeFilter` function:

```
if(params.includes('location')){
     remove.push('coords')
     remove.push('r')
 }
```
Which is gross :vomiting_face: 

 I think having an "input definition" that specifies not only the default values but any other params that should be cleared when a filter is cleared removes that kind of logic.

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -1,91 +1,76 @@
...
+            return exclude
+        },
+        shouldDoCommuteSearch() {
+            return !blank(this.input.coords) && !blank(this.input.commuteLocation)
         },
-        newSearch() {
-            this.input.page = 1
-            this.isCommuteSearch = !this.isSolr && (this.input.coords && this.input.commuteLocation)
-            if(this.isCommuteSearch){
+        beforeSearch() {
```
surgiie, line 70: Instead of overwritting `newSearch` and duplicating some logic, providing a `beforeSearch` function for providers to hook into feels better imo.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -39,10 +34,20 @@ export default  {
             featuredJobs: [],
             appliedFilters: [],
             isCommuteSearch: false,
-            input: this.getInputDefaults(),
+            input: this.defaultInput,
         }
     },
-    computed:{
+    computed: {
+        inputDefinition() {
```
surgiie, line 41: Making these computed properties so they are cached and not merged/calculated as they are not/likely going to change.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -87,137 +112,148 @@ export default  {
...
 
         if(!blank(this.$route.params.location)){
             this.input.location =  displayLocationFromSlug(this.input.location)
         }
 
-        if(this.searchOnLoad) {
+        if (this.searchOnLoad) {
             this.search()
+        }else{
+            this.appliedFilters = this.getAppliedFilters()
```
surgiie, line 128: Hit this issue with Zarif. On the job detail page of boeings, filter chips are available:

https://boeingveterans.directemployers.works/st-louis-mo/senior-software-engineers/FA23CED3CC6447E59F4C09777109EA0A/job/


but our code only calculates applied filters when search is done, so applied filters should be calculated when the `search-on-load` is set to false for initial loads/job detail pages.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -87,137 +112,148 @@ export default  {
...
+        setInput(filter) {
+            this.newSearch(
+                this.mergeWithDefaultInput({
+                    ...this.input,
+                    ...filter,
+                })
+            )
         },
-
-        selectPage(page) {
```
surgiie, line 176: Instead of having a very specific function for updating input and doing a search this can be done by the `setInput` function that accepts an object payload to be merged which is why i changed how pagination component emits page selection. This allows `setInput` to be generic and work for several sub components.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -87,137 +112,148 @@ export default  {
...
-            this.input["page"] = page
-            this.pushPayload()
+        filterInput(input) {
+            let excluded = this.excludePayload()
+            return omitBy(input, (v, k) => {
+                return blank(v) || excluded.includes(k)
+            })
         },
-
-        pushPayload(payload = null) {
```
surgiie, line 181: We have have been using `pushPayload` and `newSearch` interchangeably and there is no reason for the 2 to exist. The same behaviour can be accomplished with a single function now just `newSearch`

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
jeremytubbs: good idea

jeremytubbs (approved):

PR #47, opened 2020-12-29
[PD-6886] load more button pagination

nate998877:
### What does this PR do?

- Adds a load more button for fetching jobs. shows all jobs on one page
- Pre fetches jobs when using load more

### notes to Reviewer

This goes with backend PR [PD-6928] and will not work properly without those changes. 

[PD-6928]: https://directemployersfoundation.atlassian.net/browse/PD-6928


src/components/AppLoadMore.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+    <nav v-if="totalPages > 1 && currentPage < totalPages">
+        <button class="button" type="button" @click="loadMore">
+            Load More
```
surgiie, line 4: May check with @jeremytubbs on this but i would honestly prefer we move away from providing styles because of how much trouble they have given Zarif. I would prefer that we just give elements some reasonable class names and let Zarif do the styling on his end.

As an example add a `load-more-pagination` class on `nav` and a `load-more-pagination__button` class on  `button`  and then just let Zarif/RR worry about how it should be styled. Again, check with @jeremytubbs or maybe we should discuss this as a team.


Side note and trivial:: Consider adding a `text` prop or wrapping `Load More` in a `<slot>` to allow implementer to customize text.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -21,31 +21,40 @@ export default {
...
             status: {
                 loading: false,
                 error: false,
             },
+            jobs: [],
+            jobDisplay: [],
             filters: [],
             pagination: {},
             featuredJobs: [],
+            isFirstLoad: true,
```
surgiie, line 36: `isFirstLoad` doesnt appear to be used anywhere.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -118,16 +128,27 @@ export default {
...
         })
 
-        if(!blank(this.$route.params.location)){
-            this.input.location =  displayLocationFromSlug(this.input.location)
+        if (!blank(this.$route.params.location)) {
+            this.input.location = displayLocationFromSlug(this.input.location)
         }
 
+        if (this.isLoadingMore) {
+            this.siteConfig.num_items = this.siteConfig.MAX_PAGE_SIZE
```
surgiie, line 136: Not sure if modifying `siteConfig` directly is a good idea. This seems like it could introduce weird bugs since site config is dependent by several components but could be wrong.

surgiie (changes requested):
Overall the code is in way better shape than the initial draft.

### Overall thoughts/Future Discussion.
* Not sure im a fan of `pagination_type` being a config option. It has a false sense that changing it will render the appropriate pagination component but it doesnt. The implementer still has to swap in the pagination component type but again maybe this is fine as its not our problem to do this for RR. I understand that is used to tell the back end what paginator type to use but you can accomplish that without a new config option. See comments i made :

https://github.com/DirectEmployers/microsites-search-api/pull/56

* Im inclined to have a single pagination component that is controlled by props. It either is a single "LoadMore" button or it has pages rendered, and then either a `loadMore` or a `pageSelected` event is emited, this allows the component to swap out without any manual intervention,  but maybe thats a bad call. :man_shrugging: 

* Though the code is alot cleaner, im still not sure im a fan of mixing logic for both infinite and  paginated jobs. 


* Without VueX its a bit hard to keep state management clean. Modifying `siteConfig` directly from props is an anti pattern, and im surprised i didnt see any console logs.

### Changes to be made.
Most comments are up to your discretion, but the following should be changed.


* When you switch back to "page" pagination and use `AppPagination` instead, the jobs dont change across pages, they stay the same no matter how many pages you navigate. It is not until you do a page reload with the current page param that you see the appropriate jobs load. I think this because of some missing splice logic on `jobDisplay`, i only saw code for this happening in `created` and not on every search. Tthis is en example result of what i mean about not liking the mix of infinite and paginated searches.

src/components/AppLoadMore.vue
jeremytubbs: Yes, we need a more specific class for "load-more-pagination" than just button and there should be an easy way to change the text.

src/config.js
```diff
@@ -4,7 +4,10 @@ const config =  {
     project_id: process.env.GRIDSOME_GOOGLE_TALENT_PROJECT_ID,
     tenant_uuid: process.env.GRIDSOME_GOOGLE_TALENT_TENANT,
     company_uuids: [process.env.GRIDSOME_GOOGLE_TALENT_COMPANY],
-    client_events: true,
+    client_events: true, // Should be extracted to a separate config at some point
+    pagination_type: "load_more", // Options: [load_more, page] load_more uses offset, page uses pages to load more jobs
```
jeremytubbs, line 8: I would prefer we keep naming consistent with the backend "LimitOffset" and "PageNumber" also change to a pagination object. The whole pagination object should be optional with defaults set on the backend.
```
pagination: {
    num_items: 15,
    type: "LimitOffset"
}
```

src/config.js
```diff
@@ -4,7 +4,10 @@ const config =  {
...
+    num_items: 15,
+    MAX_PAGE_SIZE: 90,
```
jeremytubbs, line 10: I don't like that we are requesting 90 jobs at a time and slowly displaying them... I would prefer this be a bit smarter and always be just one page request ahead. For example first request would take 30 jobs, each press of the "Load More" button would request 15 more and display the 15 it has already loaded and not displayed.

jeremytubbs (changes requested):

src/pages/jobs.vue
```diff
@@ -45,12 +47,14 @@
                                 <div class="text-2xl">
                                     {{ pagination.total }} jobs found
                                 </div>
-                                <AppPagination
-                                    @pageSelected="setInput"
-                                    :current-page="pagination.page"
-                                    :total-records="pagination.total"
-                                    :total-pages="pagination.total_pages"
+                                <AppLoadMore
```
jeremytubbs, line 50: We can easily switch these out if we wanted to do this.
 
```
<AppLoadMore v-if="$siteConfig.pagination_type == 'load_more'" />
<AppPagination v-else />
```
```
components: {
        AppLoadMore: () => import("~/components/AppLoadMore"),
        AppPagination: () => import("~/components/AppPagination"),
}
```

src/config.js
nate998877: This decreases the number of requests that need to be made. The logic is the same if we set it to 30. Is there a specific reason we don't want to fetch more than 1 page?

src/config.js
nate998877: Currently, a default page size needs to be set on the frontend in order to determine how many jobs to display when using with load more. For the time being, I've removed the default page size on the frontend. Should that be added back? Is there a specific reason we can't have it as a requirement for the config?

src/components/Search/Providers/mixins/provider.js
```diff
@@ -200,31 +239,40 @@ export default {
...
+                .concat(this.applyFilters())
         },
         search() {
+            if (!this.isLoadingMore || this.jobDisplay.length == 0) {
+                this.status.loading = true
+            }
             this.beforeSearch()
-            this.status.loading = true
-            this.service(this.filterInput(this.input), this.siteConfig)
+            return this.service(this.filterInput(this.input), this.siteConfigClone)
```
jeremytubbs, line 256: Do not clone the site config but instead spread offset pagination data with the input data.

src/config.js
```diff
@@ -4,7 +4,12 @@ const config =  {
     project_id: process.env.GRIDSOME_GOOGLE_TALENT_PROJECT_ID,
     tenant_uuid: process.env.GRIDSOME_GOOGLE_TALENT_TENANT,
     company_uuids: [process.env.GRIDSOME_GOOGLE_TALENT_COMPANY],
-    client_events: true,
+    client_events: true, // Should be extracted to a separate config at some point
+    pagination_type: {
+        type: "load_more", // Options: [load_more, page] load_more uses offset, page uses pages to load more jobs
+        num_items: 15,
+        max_page_size: 90, //should be a multiple of num_items and under 100
+    },
```
jeremytubbs, line 12: remove `pagination_type` and put it back to being a prop on the Search Provider. The search provider will add `offset` and optionally `num_items` to the data. If offset is present it will use the OffsetPaginator.  Remove `max_page_size` and just request a new offset each time the Load More button is clicked.

jeremytubbs (commented):
I also found a bug in offset not being calculated properly on subsequent requests causing duplicate jobs to be rendered. Comments below further simplify the requirements of this task and should make this bug a non-issue. 
- remove max_page_size
- remove pagaination_type from site config
- submit offset as part of data to the search but not part of input

PR #49, opened 2021-01-04
[PD-6886] load more button pagination

nate998877:
RE-Opened PR due to mess up refer to https://github.com/DirectEmployers/microsite-js/pull/47 for progress prior to this PR.


jeremytubbs (changes requested):
There is a problem with the initial request after doing a filter. To reproduce load page http://localhost:8080/jobs?sort=relevance then select any filter you should see only 7 of the results are displayed.

jeremytubbs:
- Frontend should hide Load More button when all jobs have been loaded.

src/components/AppLoadMore.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
surgiie, line 1: Thinking about his some more, this component feels very unnecessary, We really dont need to have a dedicated component that is simply a button in my opinon. We can accomplish the same functionality with the available data in the slot.

We already pass `loadMore` function to the slot so a button that calls this accomplishes what the component does without adding more components on our end.


```html
<nav  v-if="totalJobs > 1 && currentJobs.length < totalJobs">
    <button type="button" @click="loadMore">
            Load More
    </button>
</nav>
```

In addition, a computed property for ` v-if="totalJobs > 1 && currentJobs.length < totalJobs"` might be nice for readability purposes `eg shouldLoadMore` but thats trivial and ill leave it at your discretion.

jeremytubbs (changes requested):
Getting an error in the console on initial page load http://localhost:8080/jobs?sort=relevance
image

nate998877:
> Getting an error in the console on initial page load http://localhost:8080/jobs?sort=relevance

I thought this error had been fixed. A misplaced .length was to blame. Fixed.

jeremytubbs (changes requested):
Offset is being miscalculated on second request causing duplicate jobs to be returned and displayed.  The second request is only doing a 15 offset after the first request takes 30 jobs.

image

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
```diff
@@ -100,7 +100,13 @@ export default {
             )
         },
         beforeSearch() {
-            if ((this.isCommuteSearch = this.shouldDoCommuteSearch())) {
+            if(this.isFirstLoad && this.isLoadMore){
+                delete this.input.page
```
surgiie, line 104: Overall, Im not a fan of code that is meant for hybrid use like this because of differences between load more and paginated provider types. This stems from my concerns of not liking the mix of load more and paginated provider logic but that entails a team discussion of what we agree on for the design of the provider. 


For now, you only need to delete page in the shared input definition in provider and as long as you make sure 

`:is-load-more="true"` prop is set on every instance of the provider on every page, you wont see bugs with page still being persisted to the url regardless of you deleting it like we saw on the home page.


Either delete it in `sharedInputDefinition` or `beforeSearch` not both.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -18,23 +18,31 @@ export default {
...
+            jobs: [],
+            jobDisplay: [],
             filters: [],
             pagination: {},
             featuredJobs: [],
+            isFirstLoad: true,
             appliedFilters: [],
             isCommuteSearch: false,
             input: this.defaultInput,
+            paginationData: {num_items: this.siteConfig.num_items},
```
surgiie, line 45: Is there any reason you couldnt reuse the existing `pagination` data ?

src/components/Search/Providers/AppGoogleTalentSearchProvider.vue
nate998877: You are correct, I believe that was changed at some point and only in sharedInputDefinitions is page being deleted.

src/components/Search/Providers/mixins/provider.js
nate998877: the existing pagination data is updated by the backend every new search. In order to keep a consistent state I used a separate property. I'll look into what's necessary to convert it to use the existing pagination property and update it if it makes sense.

surgiie: Or can you add the data you need in `this.input` ? 

This would reduce the object merging you we are doing on line 266:

`return this.service({...this.filterInput(this.input), ...this.paginationData} ...`

src/scss/components/load-more.scss
```diff
@@ -0,0 +1,10 @@
+.load-more-pagination-button {
```
surgiie, line 1: Again, not sure we should have this component and its associated styles.

If we really need to keep this for now, make these vanilla css and try to use BEM convention. 

I would prefer this file didnt exist but we can discuss with @jeremytubbs

src/components/Search/Providers/mixins/provider.js
nate998877: Adding the pagination data to this.input adds it to the URL, which I'm trying to avoid.

src/scss/components/load-more.scss
nate998877: This is an @jeremytubbs discussion. I can convert it to BEM. Are we adopting BEM as our convention/have we?

src/components/Search/Providers/mixins/provider.js
nate998877: Back with some updates on using pagination vs paginationData. I don't think it makes sense to try and use pagination to handle the state that paginationData does, it could be renamed to something better, but I believe it should remain separate.

There are a couple of reasons.
- pagination is updated every search by the backend, so updating pagination would have to happen after a search instead of before, which I believe would be confusing and could be difficult to handle if a search is performed that doesn't impact pagination. (not sure that we're doing any, but similar jobs comes to mind)
- Pagination has several properties that are irrelevant to fetching ie total, has_more_pages
- The returned pagination data is from the last request, which is potentially outdated for the next request. IE P.offset = 0 when PD.offset = 30.

I'm open to discussion about it though, would just like some reasoning.

src/scss/components/load-more.scss
surgiie: Personally dont like BEM but thats how the other components are named so i only suggested that for consistency.

surgiie:
The following is my opinion on how i think this should work. This will ultimately be a team discussion and i just want my opinion documented here for future reference  for said  discussion.


Although i think the prefetching is nice, i personally dont think we should care that much for this type of response. I would think most users will click "Load More"  twice maybe three times before they refine their search?  In addition we are removing `page` parameter to accomplish something we could do with `page` parameter. Originally i thought this was required when we were calculating an offset to submit instead, but after playing around with the code, i dont think this is necessary.

Trying to do prefetching and removing page parameter adds several edge case issues that could just be avoided.

As an example, we added the following checks: 

* Check if its the first or not first load
* Modify `num_items` conditionally which added a 2nd `paginationData` object.
* Delete `page` out of the payload/url.


In my opinion, this could work alot simpler and work with the `page` param still present.

This should just work by concatenating the jobs on every response so you have all the jobs as a reference to determine what is being displayed. Then display the needed jobs by splicing 0 to the total number of  number of collected jobs.

So the search function would look something like this:


```js
search() {
    this.status.loading = true

    this.beforeSearch()

    return this.service({...this.filterInput(this.input)}, this.siteConfig)
        .then(resp => {
            const data = resp.data || {}
            this.featuredJobs = data.featured_jobs || []
            this.pagination = data.pagination || {}
            this.filters = data.filters || {}
            let responseJobs = data.jobs || []
           
            if(!this.isLoadMore){
                this.jobs = responseJobs
                this.jobDisplay = this.jobs
            }else{
                this.jobs = this.jobs.concat(responseJobs)
                this.jobDisplay = clone(this.jobs).splice(0, this.siteConfig.num_items * this.input.page)
            }

            this.meta = data.meta || {
                source: SOLR,
            } //prevents sites from erroring when unable to connect to api
            this.appliedFilters = this.getAppliedFilters()
            this.searchCompleted(data)
        })
        .catch(err => {
            this.status.error = err
        })
        .finally(() => {
            this.status.loading = false
        })
}

```

Why I prefer this:


* No more specific checks about page or which load it is
* No more having to modify num items conditionally based on which load were in
* We get rid of the offset PR in search api and accomplish both scenarios with a single `page` parameter. There is no need to try to calculate an offset anymore.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -219,16 +259,20 @@ export default {
                 .concat(this.applyFilters())
         },
         search() {
+            if (!this.isLoadMore || this.jobDisplay.length == 0) {
```
surgiie, line 262: I dont think we should be preventing loading state based on provider type.

Loading should always be toggled to true. If its a matter of not showing "Loading..." in the UI then we should just be removing the loader component when the provider type is "Load More"

src/scss/components/load-more.scss
```diff
@@ -0,0 +1,10 @@
+.load-more-pagination__button {
+    @apply transition-colors duration-100 inline-block;
```
jeremytubbs, line 2: unfortunately, we are no longer writing Scss or using tailwind... so you need to convert this to normal css.

jeremytubbs (changes requested):
AppLoader now shows every time the load more button is clicked. 
- do not show loading state overlay when clicking the load more button

jeremytubbs:
Closing PR now so that we know it is ready for review when it is open again.

nate998877:
I believe the PR is ready.

src/pages/jobs.vue
```diff
@@ -1,16 +1,19 @@
...
             <template v-slot="{
                 jobs,
                 sort,
                 input,
                 source,
                 status,
                 hasJobs,
+                loadMore,
                 setInput,
+                setFilter,
```
jeremytubbs, line 13: My IDE is showing `setFilter` and `setInput` as not used.

src/pages/jobs.vue
```diff
@@ -19,7 +22,7 @@
                 isCommuteSearch,
                 getFilterOptions,
             }">
-                <AppLoader v-if="status.loading"/>
+                <AppLoader v-if="status.loading && !isLoadMore || status.loading && !jobs.length"/>
```
jeremytubbs, line 25: `status.loading && !isLoadMore ||` can be removed and `isLoadMore` can be removed from the slot data from the provider.

src/pages/jobs.vue
nate998877: They are unused by this component, but other pagination types do use them. Every time a pagination component is switched should the unused slot data be removed even if it's likely to be used later on?

src/pages/jobs.vue
nate998877: You are correct, This has been updated.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -1,4 +1,4 @@
-import {omitBy, clone, merge, startCase, uniqBy} from "lodash"
+import {omitBy, clone, cloneDeep, startCase, uniqBy} from "lodash"
```
jeremytubbs, line 1: cloneDeep is no longer being used

jeremytubbs (approved):
Approving but definitely waiting for @surgiie since this is also my code now.

src/components/Search/Providers/mixins/provider.js
```diff
@@ -194,23 +221,35 @@ export default {
         },
 
         search() {
-            this.beforeSearch()
             this.status.loading = true
-            this.service(this.filterInput(this.input), this.siteConfig)
+            this.beforeSearch()
+            if (this.isLoadMore) {
+                this.beforeLoadMoreSearch()
```
surgiie, line 227: I like that the math stuff is now in a  `beforeLoadMoreSearch` function, but the code in it right now would not allow any sub provider component to overwrite without duplicating the logic. I think its fine right now but maybe we should consider moving this into the `loadMore` function and leave `beforeLoadMoreSearch` for the individual provider components to extend as needed. Again i don think it's really necessary as i cant think of a reason why solr or google talent providers would need this.

surgiie (changes requested):
When clicking on a filter `page` is inserted back in the url because the `AppSearchFilter` component resets it to one when a filter is clicked.

Add a `resetPage` boolean prop or something of that nature that determines if the search filter should be resetting page.

surgiie (approved):
LGTM! :tada:

PR #52, opened 2021-01-13
[PD-6935] filter chip as link

nate998877:
###What does this PR do?
Simple PR to change app chips from emitting event to use g-links for SEO purposes

##Notes to Reviewer
The g-link wraps the slot in app chip. I have a feeling this isn't preferred.
Due to the simplicity of any necessary changes in this regard. I still believe it's worthwhile to open the PR at this point.

### DIFF
--- a/src/components/AppChip.vue
+++ b/src/components/AppChip.vue
@@ -1,12 +1,15 @@
 <template>
-    <span class="chip" tabindex="0" @click="clickedChip">
-        <slot>
-            {{ text }}
-        </slot>
+    <span class="chip" tabindex="0">
+        <g-link :to="url">
+            <slot>
+                {{ text }}
+            </slot>
+        </g-link>
     </span>
 </template>
 
 <script>
+import buildUrl from "axios/lib/helpers/buildURL"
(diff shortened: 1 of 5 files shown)


src/components/AppChip.vue
```diff
@@ -24,26 +27,37 @@ export default {
...
             if(e.keyCode == ENTER_KEY && document.activeElement == this.$el){
                 this.clickedChip()
             }
         },
-        clickedChip(){
-            this.$emit('chipClicked', this.name)
+
+        removeFilter(){
+            if(this.name === "*"){
+                return buildUrl(this.$route.path)
```
surgiie, line 55: I think you can just return `this.$route.path` here. `buildUrl` is only useful when you want a query string attached to the url using an object `e.g buildUrl(path, {....})`

jeremytubbs (approved):
Nice work! I like how this moves the `removeFilter` logic out of the search provider.

surgiie:
> Nice work! I like how this moves the `removeFilter` logic out of the search provider.

Same. LGTM!

surgiie (approved):

PR #54, opened 2021-01-15
Fix filter chip clears bug.

nate998877:
Fix broken links with chips

### DIFF
--- a/src/components/Search/Providers/mixins/provider.js
+++ b/src/components/Search/Providers/mixins/provider.js
@@ -212,7 +212,7 @@ export default {
                     return !blank(this.input[filter.name])
                 })
                 .map(filter => {
-                    let clears = this.inputDefinition[filter.name].clears || []
+                    let clears = this.inputDefinition?.[filter.name]?.clears || []
                     return {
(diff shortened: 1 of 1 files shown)


src/components/Search/Providers/mixins/provider.js
```diff
@@ -212,9 +212,10 @@ export default {
                     return !blank(this.input[filter.name])
                 })
                 .map(filter => {
+                    let clears = this.inputDefinition[filter.name].clears || {}
```
surgiie, line 215: The default should be an array, not an object. ` this.inputDefinition[filter.name].clear || []`

surgiie (changes requested):

src/components/Search/Providers/mixins/provider.js
surgiie: You may also consider doing a default for the definition object in case that is not present ?
`(this.inputDefinition[filter.name] || {}).clears || []`

surgiie (approved):

jeremytubbs (approved):

PR #57, opened 2021-01-20
remove source calculation from title autocomplete

surgiie:
We are removing google title complete and are using solr for all autocompletes. This code removes trying to calculate the endpoint.


src/services/search.js
```diff
@@ -54,18 +54,11 @@ export function commuteSearchService(input, siteConfig){
 
 
 export class TitleCompleteService {
```
nate998877, line 56: Title autocomplete looks good, but location autocomplete's URL also needs updated. Currently, location autocomplete 404s.

nate998877 (approved):

PR #58, opened 2021-01-21
Retry is gross

nate998877:
### What does it do?
Removes the retry method from Helpers and appgooglelocationautocomplete.

### Notes To Reviewer
I don't think the code would handle things well if google refused to load. Consider wrapping is try/catch. Although onload may never trigger?


src/components/Search/AppGoogleLocationAutocomplete.vue
```diff
@@ -24,24 +23,20 @@ export default {
...
     computed: {
         apiScriptUrl() {
-            let key = this.apiKey
-
-            return `https://maps.googleapis.com/maps/api/js?key=${key}&libraries=places`
+            return `https://maps.googleapis.com/maps/api/js?key=${this.apiKey}&libraries=places`
         },
         placesApiLoaded() {
-            return typeof get(window, "google.maps.places") == "object"
+            return window?.google?.maps?.places
```
surgiie, line 31: This returns `undefined` on the inital load or an object  when it is loaded, the name implies that it returns a boolean determining if the places api has loaded, either rename the computed property or return a boolean.

surgiie (changes requested):
This only works on the initial load of the autocomplete when you first open "Commute Search" modal. If you do  a commute search and reopen the modal and try to change your commute location, it doesnt autocomplete anymore.

This is because you need to reinit the autocomplete during the `updated` event:

```js
updated(){
    this.initAutocomplete()
}
```

src/components/Search/AppGoogleLocationAutocomplete.vue
```diff
@@ -10,7 +10,6 @@
 
 <script>
 import {get} from "lodash"
```
jeremytubbs, line 12: unused import

jeremytubbs (changes requested):
Can we rename this component to `AppGooglePlacesAutocomplete` since it is using the places api?

Pre-existing Issues (need to check out QC to confirm)
- changing input without selecting from the autocomplete dropdown does not clear the commute form coords
- commute search only works once per hard page refresh
     - after doing a commute search then reopening the modal to do another commute search the Google Places API autocomplete does not work
     - after doing a commute search and hitting clear all the and attempting to do another commute search the Places API autocomplete does not work

Improvements
- form submit button should be disabled if the user has not selected from the autocomplete and coords are available to do the search

surgiie (approved):
Approving with one change to make per Jeremy's feeback.

Coords are not cleared when the user is typing in the autocomplete input in commute search form.

You can fix this by adding `@input="input.coords = ''"` directly on the `AppGooglePlacesAutocompleteInput` in the commute search form.

You also have conflicts/need to merge latest QC.

nate998877:
> Approving with one change to make per Jeremy's feeback.
> 
> Coords are not cleared when the user is typing in the autocomplete input in commute search form.
> 
> You can fix this by adding `@input="input.coords = ''"` directly on the `AppGooglePlacesAutocompleteInput` in the commute search form.
> 
> You also have conflicts/need to merge latest QC.

I can add your recommended fix, but I still cannot replicate this bug, which is frustrating.

surgiie:
> > Approving with one change to make per Jeremy's feeback.
> > Coords are not cleared when the user is typing in the autocomplete input in commute search form.
> > You can fix this by adding `@input="input.coords = ''"` directly on the `AppGooglePlacesAutocompleteInput` in the commute search form.
> > You also have conflicts/need to merge latest QC.
> 
> I can add your recommended fix, but I still cannot replicate this bug, which is frustrating.

Its not really a visual bug, but you can see it in the vue dev tools. When you type in commute search location input, the coords on the "input" in provider  do not clear.

jeremytubbs (approved):

PR #60, opened 2021-01-25
[PD-6885] two button pagination

nate998877:
### What does this PR do?
This adds another pagination component that uses a simpler 2 button design


src/components/AppTwoButtonPagination.vue
```diff
@@ -0,0 +1,58 @@
+<template>
```
surgiie, line 1: I prefer that this component was called `AppSimplePagination` 

We also now have 3 pagination components, I would also prefer we group these into a `Pagination` directory to add better organization to the the components directory.

src/components/AppTwoButtonPagination.vue
```diff
@@ -0,0 +1,58 @@
...
+    <nav aria-label="pagination" class="simple-pagination">
+        <button
+            class="simple-pagination-button"
+            v-if="currentPage > 1"
+            @click.prevent="selectPage(previousPage)"
+        >
+            Back
+        </button>
+        page {{currentPage}}
```
surgiie, line 10: I dont think we should worry about showing the current page.

I would let the implementer decide with a slot:

```html
<slot :current-page="currentPage"></slot>
```

src/components/AppTwoButtonPagination.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+    <nav aria-label="pagination" class="simple-pagination">
+        <button
+            class="simple-pagination-button"
+            v-if="currentPage > 1"
+            @click.prevent="selectPage(previousPage)"
+        >
+            Back
```
surgiie, line 8: Adds slots to allow customization of the `Back` and `Next` text via slots. This will allow people to use custom text or arrow icons as needed.

surgiie (changes requested):
Looks good for the most part but I made a few comments for changes.

src/components/AppTwoButtonPagination.vue
nate998877: I will leave moving things until after Jeremy's PR on that. Have renamed though.

src/components/AppTwoButtonPagination.vue
jeremytubbs: I agree.. but I would just make a default empty slot `<slot />`

src/components/AppTwoButtonPagination.vue
jeremytubbs: Please go ahead a move the pagination components into a folder.

jeremytubbs (changes requested):
- remove `page #` and add an empty slot
- move the three components into a pagination folder

surgiie:
@nate998877 Just realized this will need documentation changes.

Can you make a PR in the docs repo, that updates docs for these changes.

https://github.com/DirectEmployers/microsite-docs

nate998877:
> @nate998877 Just realized this will need documentation changes.
> 
> Can you make a PR in the docs repo, that updates docs for these changes.
> 
> https://github.com/DirectEmployers/microsite-docs

It's been a low priority item for me. I'll get on that now!

surgiie (approved):

PR #61, opened 2021-01-25
Generic fetch component.

surgiie:
* Adds a generic fetch component by extracting alot of the useful code from job fetch component
* Adds `Fetch` folder to group fetching components.
* Adds `tag` prop on `AppJobFetch` component. 

### Why

At some point RR will start integrating with other APIs and will be "fetching" data from other places. I think having a couple specific fetching components like `AppJobFetch` is nice, but we wont be able to cover every fetch edge case, so having a generic fetch component to make this seamless will be super useful:

```html
<AppFetch endpoint="/some/endpoint" :options="{ headers: {....}}">
   <template v-slot="{ data, status }">
        <!-- render data or show loader with status-->
   </template>
</AppFetch>
```

The `endpoint` prop can also be a promise or  function that returns a promise. In the `AppJobFetch` component that is the `getJob` helper function we have.  This is to allow flexibility on what does the "fetching". 

You can hook into success and error events using  the `onResolve` and `onError` props as needed:

```html
<AppFetch endpoint="/some/endpoint" :on-resolve="requestCompleted" :on-error="requestFailed">
   <template v-slot="{ data, status }">
        <div v-if="data">....</div>
   </template>
</AppFetch>
<script>
export default{
     methods:{
         requestCompleted(data, response, setStatus){
                // do something with data and mark statuses as needed:
                // e.g setStatus({ resolved: true})
         },
         
         requestFailed(error, setStatus){
              // do something and mark status as needed:
              // e.g setStatus({ error, resolved: true})
         }
     }
}
</script>
```
### Other Future Considerations

* Wrapper form component around axios. This is to ease api integrations like JobVite in 1.0. It was not pretty or easy to work with.


src/components/Fetch/AppFetch.vue
```diff
@@ -0,0 +1,124 @@
...
+            return { params: this.parameters, ...this.options }
+        },
+        requestHandler() {
+            if (typeof this.endpoint == "function") {
+                return this.endpoint(this.requestPayload)
+            }else if(this.endpoint instanceof Promise){
+                return this.endpoint
+            }
+            return axios.get(this.endpoint, this.requestPayload)
+        },
```
nate998877, line 86: This method gives me bad juju. I don't have any great insight into how to improve it though.

 The only thing I can think of is changing 

> typeof this.endpoint == "function" 

to 
> this.endpoint instanceof Function 

to keep consistency.

nate998877 (approved):

nate998877 (commented):
Fix the merge conflicts and I'm ready to approve. The comments I've added are opinions and don't influence my acceptance of the PR.

PR #66, opened 2021-02-08
allow redirects for /guid for rss urls

surgiie:
## What does this PR do?
* Adds a rule/page for redirecting `/:guid` routes to job detail pages.
* Changes job fetch component to parse route guid param for view source. RSS feed url guids contain the guid + viewsource. So this component will now parse guid and view source when redirecting from `/:guid` to `/:title/:location/:guid?vs=parsedVS...`

I also snuck in a new `delayLoadBy` prop on the provider to allow a delay on toggling loader when searches are done. This is for Zarif and the experimenting we are doing with the content loader. The content loads so quickly that the content loader just does a weird flicker and doesnt look good. Ive tested a delay of half a second and it looks alot better so sneaking this small unrelated change in. 

## Testing

Grab any guid from results and try navigating to:

`/:guid/` and you will be redirected to `/:title/:location/:guid?vs=...`

Example:

http://localhost:8080/DD4437D149E546A892F76DA93D76A92B9000

Should redirect you to:

http://localhost:8080/moorhead-mn/patient-access-representative/DD4437D149E546A892F76DA93D76A92B/job/?vs=9000


src/components/Fetch/AppJobFetch.vue
```diff
@@ -1,16 +1,18 @@
...
-                :status="status"
-                :job="job"
-            ></slot>
+    <AppFetch
+        :endpoint="fetchJob"
+        ref="fetch"
+        :tag="tag"
+        :on-resolve="resolveJob"
+    >
+        <template v-slot="{status}">
```
jeremytubbs, line 8: why did you change this from being a default slot?

jeremytubbs (approved):

PR #68, opened 2021-02-09
[PD-6974] - keep old url params out of input

jeremytubbs:
## What does this PR do?
- updates the vet-jobs URL
- Adds params defined as filters to the input and adds params not in the input to extra data.

### DIFF
--- a/src/components/Search/Providers/mixins/provider.js
+++ b/src/components/Search/Providers/mixins/provider.js
@@ -258,10 +258,7 @@ export default {
             if (this.isLoadMore) {
                 this.beforeLoadMoreSearch()
             }
-            return this.service({
-                    ...this.filterInput(this.input),
-                    ...this.getExtraData()
-                }, this.siteConfig)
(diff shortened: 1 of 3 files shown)


src/components/Search/Providers/mixins/provider.js
```diff
@@ -205,13 +205,39 @@ export default {
...
-
+        getUrlFiltersObject(object) {
+            for (const key in object) {
+                if (! this.getConfigFilterSlugs().includes(key)) {
+                    delete object[key]
+                }
+            }
+            return object
+        },
+        getUrlExtraDataObject(object) {
```
nate998877, line 230: This method is unused?

nate998877 (approved):

src/components/Search/Providers/mixins/provider.js
jeremytubbs: It is used in `getExtraData` line 211

PR #70, opened 2021-02-10
[PD-6856] old to new redirect

nate998877:
## What does this PR do?
This PR adds support for using "old" format filter urls.

## Notes to Reviewer
This whole thing feels really janky and while it does appear to work appropriately. I would imagine it has the potential to introduce lots of weird edge cases. Test thoroughly!

## Associated backend PR
https://github.com/DirectEmployers/microsites-search-api/pull/74


src/components/Search/Providers/mixins/provider.js
```diff
@@ -300,12 +298,23 @@ export default {
             return options
         },
         setInput(filter) {
-            this.newSearch(
-                this.mergeWithDefaultInput({
+            if(this.getUrlExtraDataObject(this.$route.params)){
```
jeremytubbs, line 301: Can you please explain these changes?

nate998877: This might need to be changed since it comes from a point of naivety with how the urls work, but when using the old url format we have to prevent route params from entering input. I don't see any instances of using url slugs with the new format, so I assumed they'd only be applicable for old urls. Thus I just completely omit them and instead handle them as extra data.

jeremytubbs (changes requested):
Can you close this PR or just make a PR with only expanded solr_queries? And the App404 change. We will work on the provider changes later.

nate998877:
PR is no longer applicable

PR #88, opened 2021-03-02
Search filter component changes to render canonical urls. 

surgiie:
## What does this PR do?
* Makes changes to the `AppSearchFilter` component to render out links as built by the backend now (associated back end PR below)
* Change the default path of `AppSearchFilterChip` to `/jobs` instead of using the current path. I hit a navigation duplicated error because of the urls now being canonical and not using query params. 
* I couldnt get rid of query params entirely because of commute search (:sadparrot:) . Those params should stay present when performing a commute search so that the commute search doesnt switch to a regular search when selecting filters. This is now handled by a `preserveQueryParams` prop which determines whether or not to do this.

## Notes to Reviewer
Associated backend PR : https://github.com/DirectEmployers/microsites-search-api/pull/82

## Document Changes
There are changes to be made to docs but we need to talk about this once this hits production. We also need a reliable way to determine what all change as we merge qc to production. Something like the RRR and DDD we have in MyJobs.

### DIFF
--- a/src/components/Search/AppSearchFilter.vue
+++ b/src/components/Search/AppSearchFilter.vue
@@ -1,9 +1,9 @@
 <template>
     <component
-        class="search-filter"
         :is="tag"
-        v-if="isVisible && hasOptions"
         v-bind="$attrs"
+        class="search-filter"
+        v-if="isVisible && hasOptions"
     >
         <h3 v-if="display" class="search-filter-display">
             Filter By {{ display }}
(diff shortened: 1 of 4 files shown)


src/components/Search/Providers/mixins/provider.js
```diff
@@ -153,16 +155,32 @@ export default {
...
+        this.input = this.unslugifyInput(this.input)
+
         if (this.searchOnLoad) {
             this.search()
         } else {
             this.appliedFilters = this.getAppliedFilters()
         }
     },
     methods: {
+        unslugifyInput(input){
```
jeremytubbs, line 167: I think we should be excluding `q` from the `unslugifyInput` because prevents me from writing query like syntax in that field.

jeremytubbs (changes requested):

PR #89, opened 2021-03-05
[PD-7051] filter chips use meta filters

nate998877:
## What does this PR do?
Changes the filter chips to use the active filters from meta.

## Notes to Reviewer
The meta filters are slugs vs query strings. Until the new-URL format is added these will be broken.

### DIFF
--- a/src/components/Search/AppSearchFilterChip.vue
+++ b/src/components/Search/AppSearchFilterChip.vue
@@ -1,6 +1,6 @@
 <template>
     <span class="search-filter-chip" tabindex="0">
-        <g-link :to="url">
+        <g-link :to="link">
             <slot>
                 {{ text }}
             </slot>
@@ -9,17 +9,16 @@
 </template>
 
 <script>
-const ENTER_KEY = 13
 
 export default {
     props: {
-        url: {
+        link: {
             type: String,
             required: true,
         },
         text: {
             type: String,
-            required: true,
+            required: false,
         },
     },
 }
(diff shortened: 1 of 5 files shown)


jeremytubbs (changes requested):
Filter chips should only be links.. the method hitEnter and the eventListeners should longer be required.

nate998877:
> Filter chips should only be links.. the method hitEnter and the eventListeners should longer be required.

I thought those might be for navigation purposes, so intentionally left them in, since I don't know.

src/components/Search/AppSearchFilterChip.vue
```diff
@@ -9,76 +9,17 @@
 </template>
 
 <script>
-import buildUrl from "axios/lib/helpers/buildURL"
 const ENTER_KEY = 13
```
jeremytubbs, line 12: const no longer needed

src/components/Search/AppSearchFilterChip.vue
```diff
@@ -9,76 +9,17 @@
...
-            query = {...query, ...input}
-
-            if (!Array.isArray(names)) {
-                names = [names]
-            }
-            for (let item of names) {
-                delete query[item]
-            }
-            return buildUrl(this.path, query)
+            required: true,
```
jeremytubbs, line 22: should be `required: false`

jeremytubbs (changes requested):
Minor changes

jeremytubbs (approved):

PR #91, opened 2021-03-09
Add guid view source prop to job mixin

surgiie:
## What does this PR do?
It was brought to my attention that apply redirect urls should have a "code" attached to the end of the guid when a view source is present. This dictates certain redirect behaviors depending on the code. This PR adds the `guid-view-source` prop to the job mixin to allow this to be specified as needed and it defaults to 10 per Seth's request. If the prop is explicitly set to `false` then nothing is attached.

### Testing

Start the server and every job apply url should have `10` attached to the end of the guid when you visit the site with `?vs=someNum` 

Play around with changing this in the job component on `AppJobDetail` e.g `:guid-view-source="9000"` or `:guid-view-source="false"`


zarifrahat (approved):
Looks good to me!

jeremytubbs (approved):

PR #92, opened 2021-03-09
[PD-7007] paginations use URL params

nate998877:
## What does this PR do?
This changes the pagination components from using buttons to links

## Notes to Reviewer
It feels like a lot of code was removed. I don't think it's used at this point, but it's worth ensuring that's the case.

## Document Changes
I don't think we document the mechanisms by which the components work do we? If so let me know and I'll add a pr.


src/components/Pagination/AppPagination.vue
```diff
@@ -110,6 +106,18 @@ export default {
                 Math.min(this.current + pageLimit, this.totalPages)
             )
         },
+
+        link(page){
+            if(typeof(page) !== 'number'){
+                return ""
+            }
+            let regex = /(\?page=)(\d+)|(&page=)(\d+)/
```
surgiie, line 114: Not sure a regex is really needed, feels a bit more complex than it has to be.
Couldnt you just utilize the `$route.query` object ? I think the function could be simplified to

```js
import buildUrl from "axios/lib/helpers/buildURL"

....

link(page){
    if(isNaN(page)){
        return ""
    }
    let query = this.$route.query
    query.page = page

    return buildUrl(this.$route.path, query)
},

```

src/components/Pagination/AppPagination.vue
```diff
@@ -31,24 +29,23 @@
...
                     {{ page }}
-                </button>
+                </g-link>
             </li>
             <li></li>
             <li class="pagination__item">
-                <button
-                    type="button"
+                <g-link
+                    :to="link(nextPage)"
```
surgiie, line 40: I would honestly make this a dynamic component because right now the `...` becomes a clickable link to `/jobs` 

I would just make it a span if the page === '...'

```
<component :is="page == '...' ? 'span': 'g-link'"
   .....
>
</component>
```

src/components/Pagination/AppSimplePagination.vue
```diff
@@ -43,13 +42,15 @@ export default {
         },
     },
     methods: {
-        selectPage(page) {
-            if (!page || isNaN(page)) {
-                return
+        link(page){
```
surgiie, line 45: Same issue here, utilize `$route.query`

src/components/Search/Providers/mixins/provider.js
```diff
@@ -300,14 +299,6 @@ export default {
             options = blank(options) || !Array.isArray(options) ? [] : options
             return options
         },
-        setInput(filter) {
```
surgiie, line 303: I think this function should remain to allow Zarif to still use it in the slot/on his end.

surgiie (changes requested):

jeremytubbs (changes requested):
Please merge quality-control... never mind I did it.

jeremytubbs (changes requested):
When on `http://localhost:8080/jobs` the pagination is not allowing me to click http://localhost:8080/jobs?page=1 and when I go to http://localhost:8080/jobs?page=3 ... page 2 is missing.

image

jeremytubbs:
Also it is allowing me to click the 1 when on `page=1`

surgiie (approved):

jeremytubbs (approved):

PR #96, opened 2021-03-16
RSS vue meta

jeremytubbs:
## What does this PR do?
- add the RSS link to the head of any page using the search provider
- allows optional passing of $static from gridsome to the search provider
- clean up metaInfo for title and description (they are automatically added my gridsome)

## Document Changes
- Add static prop to Search Provider


jeremytubbs:
bugs

surgiie (approved):
I think long term as Zarif becomes more responsible of using `searchService` himself, we can just give him a function that he can put in `AppSeachForm` in his apps `e.g addRssMeta(...)` that way we dont have to worry about what pages we are on, the url will be added automatically anywhere he is using his form component.

nate998877 (approved):

PR #114, opened 2021-04-14
remove client only from fetch component

surgiie:
## What does this PR do?
Removes `<ClientOnly>` from the fetch component. This currently causes Zarif to see weird flickers on front door jobs where the footer shows first before the job detail place holder divs. This is because of the `v-if` on the slot content is added dynamically and the footer is somehow loaded before. This gives Zarif more control on doing a v-show on a placeholder div/loader followed by a client only around the actual content. 
This does require Zarif to be responsible to add `<ClientOnly>` on the job template, here's a valid example:

```html
<AppJobFetch :s3-folder="$siteConfig.s3Folder">
    <template v-slot="{job, status: {error, pending, resolved}}">
        <AppLoader v-show="pending" /> <!-- This should be v-show outside client only -->
        <!-- Then the client only content doesnt cause odd flickers-->
        <ClientOnly>
            <App404 v-if="error && !resolved" />
            <AppJobExpired v-else-if="isExpired(job)" :job="job" />
            <AppJobDetail v-else-if="job" :job="job" />
        </ClientOnly>
    </template>
</AppJobFetch>
```


jeremytubbs (approved):
I think in the future (now that QC is stable with production) we should make PRs like this against QC and then do an immediate PR and merge against production.

zarifrahat (approved):

PR #125, opened 2021-05-03
dont allow search query param to overwrite filter url

surgiie:
## What does this PR do?

Example if you click on a state filter of "New Mexico" it searches with this url:

http://localhost:8080/locations/new-mexico/jobs?page=1&sort=relevance

Then look at the city filters urls, you will see that other location filters use the current location param value:

http://localhost:8080/locations/albuquerque-nm/jobs?page=1&sort=relevance&location=New%20Mexico 

Clicking on this would use "New Mexico" as the location instead of Albuquerque New Mexico in the url which is not what the user clicked on. 

Also filters out empty params when building url.


jeremytubbs (commented):
I am still getting result urls like this - http://localhost:8080/locations/california/shift/day/jobs?page=1&sort=relevance&location=California with double location.

image

jeremytubbs:
it can get even worse - http://localhost:8080/locations/united-states/shift/day-day/jobs?page=1&sort=relevance&shift%5B%5D%5B%5D=day&shift%5B%5D%5B%5D=day&shift=day-day

nate998877 (approved):

jeremytubbs (approved):

PR #129, opened 2021-05-06
Allow hex fill on icons.

surgiie:
## What does this PR do?
Helping Zarif with PSD and being able to set the fill on these would save him some custom css.


src/components/Icons/AppHamburgerMenuIcon.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
         <title>{{ title }}</title>
-        <path d="M0 3h20v2H0V3zm0 6h20v2H0V9zm0 6h20v2H0v-2z" />
+        <path :fill="fill" d="M0 3h20v2H0V3zm0 6h20v2H0V9zm0 6h20v2H0v-2z" />
```
jeremytubbs, line 4: You could just add `class="fill-current"` to the svg and then on the top level you can declare the color. No props required. 

`<AppHamburgerMenuIcon class="text-red-700" />`

surgiie: Even better, updated.

jeremytubbs (approved):

nate998877 (approved):

PR #130, opened 2021-05-07
[PD-7025] solr endpoint services

nate998877:
## What does this PR do?
Adds services for solr endpoints

## Notes to Reviewer
Doesn't use new service, just introduces them and refactors services. You can test them by changing searchService in provider to filterSearchService or jobSearchService


src/services/search.js
```diff
@@ -29,14 +29,20 @@ export function api() {
...
-export function searchService(input, config) {
+function apiService(input, config, path) {
     const source = kebabCase(config.source)
-    const endpoint = `${source}/search`
-    if (!process.isClient || (isDevelopment() && window.location.hostname == 'localhost')){
-        return api().post(endpoint, {
-            data: input,
-            config: config,
+    const endpoint = `${source}/${path}`
+    if (process.isClient) {
```
surgiie, line 35: Rework this back to what it was or builds fail:

https://github.com/DirectEmployers/microsite-js/pull/128

surgiie (changes requested):

src/services/search.js
```diff
@@ -29,22 +29,32 @@ export function api() {
...
     const source = kebabCase(config.source)
-    const endpoint = `${source}/search`
-    if (!process.isClient || (isDevelopment() && window.location.hostname == 'localhost')){
-        return api().post(endpoint, {
-            data: input,
-            config: config,
-        })
-    }
+    const endpoint = `${source}/${path}`
+    if (!process.isClient || (isDevelopment()) && window.location.hostname == 'localhost') {
```
surgiie, line 35: The if is still off a little bit:

It shoule be:

`if (!process.isClient || (isDevelopment() && window.location.hostname == 'localhost')){`

The close parenthesis `)` on `isDevelopment` belongs at the end of  of the if

surgiie (changes requested):

src/services/search.js
```diff
@@ -29,22 +29,32 @@ export function api() {
...
+}
+
+export function jobsSearchService(input, config) {
+    return apiService(input, config, 'jobs')
+}
+
+export function filtersSearchService(input, config) {
+    return apiService(input, config, 'filters')
+}
 export function commuteSearchService(input, config) {
```
surgiie, line 58: Pretty sure `commuteSearchService` can be made generic to use `apiService` as well

surgiie (changes requested):

src/services/search.js
```diff
@@ -29,35 +29,35 @@ export function api() {
...
-        return api().post(endpoint, {
-            data: input,
-            config: config,
-        })
-    }
-    input.origin = window.location.hostname
-    return api().get(endpoint, {
-        params: input,
-    })
+export function jobsSearchService(input, config) {
```
jeremytubbs, line 51: make these singular `jobSearchService` and `filterSearchService`

src/services/search.js
```diff
@@ -29,35 +29,35 @@ export function api() {
...
+function apiService(input, config, path, path_is_endpoint=false) {
     const source = kebabCase(config.source)
-    const endpoint = `${source}/search`
-    if (!process.isClient || (isDevelopment() && window.location.hostname == 'localhost')){
-        return api().post(endpoint, {
-            data: input,
-            config: config,
-        })
-    }
+    const endpoint = path_is_endpoint ? path : `${source}/${path}`
```
surgiie, line 34: Imo id rather see us move to accepting the endpoint instead of adding this new parameter


```js
export function searchService(input, config) {
    const source = kebabCase(config.source)
    return apiService(input, config, `${source}/${path}`)
}

```

surgiie (changes requested):

jeremytubbs (changes requested):
Could you update the config to use a solr config?

src/services/search.js
jeremytubbs: just move the path calculation back the the individual services

jeremytubbs (approved):

jeremytubbs (approved):

surgiie (approved):

PR #131, opened 2021-05-14
[PD-7174] New URL routes bugfix

jeremytubbs:
## What does this PR do?
- Fixes bad recursion. Something was very wrong with how it was previously getting the filter URL paths, if you compare the amount that it was returning and the amount being stored in `rewriteRules.js` you will see that it was duplicating many URL and depending on `createPage` to filter out the duplicate URLs.  

- Removes unused default filter paths, a default filter must be defined in the config to be used in the URLs (OldSeoPaths still contain all possible old paths)

- Stops using kabab case and moves slugify method to `slugify.js` file because we cannot use ES6 imports in this file (the code is currently duplicated) 

### Release notes
RRR default filter urls will not be available unless defined in the config filters

## Notes to Reviewer
Test by running build and comparing the results of `rewriteRules.js` or adding console.logs to `grdsome.server.js` for `filterPaths.length()` also test URLs in the browser to make sure expected URLs are still working as before.

## Document Changes
No changes still uses `getFilterPaths(conf.filters)` in gridsome.serve.js

### DIFF
--- a/src/services/gridsomeFilterPaths.js
+++ b/src/services/gridsomeFilterPaths.js
@@ -7,7 +7,7 @@ const defaultFilterNames = _.map(defaultUrlFilters, "name")
 function getFilterPaths(filters) {
     let filterPaths = []
     let allFilters = getAllFilters(filters)
-    const filterGroups = getSubArrays(allFilters.reverse())
+    const filterGroups = getFilterGroups(allFilters.reverse())
 
     for (let i = 0; i < filterGroups.length; i++) {
         let path = ""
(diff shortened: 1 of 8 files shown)


src/services/slugify.js
```diff
@@ -0,0 +1,11 @@
+const _ = require("lodash")
+
+function slugify(string){
```
surgiie, line 3: There is already an existing `slugify` in `services/helpers.js`

surgiie (changes requested):

surgiie (approved):

src/services/gridsomeFilterPaths.js
```diff
@@ -1,38 +1,70 @@
...
+    let filterPaths = []
+    let allFilters = getAllFilters(filters)
+    const filterGroups = getFilterGroups(allFilters.reverse())
+
+    for (let i = 0; i < filterGroups.length; i++) {
+        let path = ""
+        for (let j = 0; j < filterGroups[i].length; j++) {
+            const param = filterGroups[i][j][0]
+            const slug = filterGroups[i][j][1]
+            path += `/${slug}/:${param}`
```
nate998877, line 17: i/j/l can be hard to differentiate. I'm not opposed to single-character counters, but for readability maybe consider distinct letters? g & f for group and filter?

nate998877 (approved):

PR #132, opened 2021-05-14
Search service simplification

nate998877:
## What does this PR do?
Simplifies the autocomplete service by calling a function for each service, similar to changes made in PD-7025.

## Notes to Reviewer
queryParams is passed with all of the requests, which it previously didn't do (only one of them used it). There doesn't seem to be any side effects of doing this as the backend ignores everything not relevant to the request. 

## Document Changes
N/A


src/services/search.js
```diff
@@ -129,11 +107,10 @@ export function parseRouteSearchInput(route) {
         clone(route.query),
         clone(route.params),
         (queryValue, paramValue) => {
-            if(paramValue && queryValue){
+            if (paramValue && queryValue) {
```
surgiie, line 110: This function has been removed in latest qc.

surgiie (changes requested):

src/services/search.js
```diff
@@ -62,61 +62,39 @@ export function commuteSearchService(input, config) {
...
-    static async get(q, queryParams = {}) {
-        let params = {
-            q: q,
-            ...queryParams,
+async function autoCompleteService(endpoint, q, queryParams = {}) {
+    let params = {
+        q,
+        ...queryParams,
+    }
+    try {
```
surgiie, line 70: I think this try catch should be removed, imo error/catch handling should be done wherever this is used.

`autocompleteService(....).catch(...)`

surgiie (changes requested):

PR #144, opened 2021-07-09
Revert "remove post request from apiService"

surgiie:
Change of plan to make testing locally easier, we will be making POST only available in local/minikube env.


src/config.js
```diff
@@ -1,3 +1,3 @@
-const config = require("./configs/frontdoor.js")
+const config = require("./configs/militaryExample.js")
```
nate998877, line 1: Can we potentially .gitignore these files or have some rule on when they can be committed?

nate998877 (approved):

jeremytubbs (approved):

PR #146, opened 2021-07-13
[PD-7257] autocomplete exact quotes

jeremytubbs:
## What does this PR do?
Adds `exactMatch` prop to autocomplete input to wrap selected value in quotes for exact match queries.


surgiie (approved):
Looks good to me.

PR #147, opened 2021-07-13
Always use production search api as endpoint unless using minikube.

surgiie:
## What does this PR do?
When testing locally, it is super confusing to see 404s on job detail pages. This is caused when testing local, the code hits search api in qc which hits qc solr and that is several days/weeks behind production so those jobs are never present and it leads to confusion. 

It also makes sense to make sure what you test locally is also what will work when you push to production.


darrint (approved):
I am very much ok with this.

jeremytubbs (approved):

PR #152, opened 2021-09-07
removed class logic

nate998877:
## What does this PR do?
removes unnecessary class logic from search services.


src/services/search.js
```diff
@@ -81,34 +81,26 @@ async function autoCompleteService(endpoint, params) {
...
 }
 
-export class TitleCompleteService {
-    static get(q, queryParams = {}) {
-        let params = {
-            q,
-            ...queryParams,
-        }
-        return autoCompleteService("v1/complete/title", params)
+export function TitleCompleteService(q, queryParams = {}) {
```
surgiie, line 84: Change these to camel case.

surgiie (changes requested):

PR #155, opened 2021-09-10
military base map component

surgiie:
## What does this PR do?
* Consolidate military base map example into a microsite-js component.

Use https://github.com/DirectEmployers/microsites-search-api/pull/266 for testing.


src/components/Search/Maps/MilitaryBaseMap.vue
```diff
@@ -0,0 +1,262 @@
...
+                .then(r => {
+                    this.bases = r.data
+                })
+                .catch(e => {
+                    this.status.error = e
+                })
+                .finally(() => {
+                    this.status.loading = false
+                })
+        },
```
nate998877, line 232: If you search for each branch individually or sort bases by branch you can avoid re-doing a search every time you toggle a branch. Only re-do the search when non-branch filter options change.

surgiie:
Closing to do some more work per PR feedback

nate998877 (approved):

jeremytubbs (changes requested):
This works the opposite of how I would expect it... I think it should be a single selection not a multiple selection. 
Currently selecting a filter option removes it from the map, I think it should select that one and remove all others from the map. We can discuss with RR today.

nate998877 (approved):

jeremytubbs (approved):

PR #159, opened 2021-09-27
[PD-7461] - job detail lighthouse

jeremytubbs:
## What does this PR do?
Component changes to be moved to site-template for job detail pages. 
- uses router methods beforeRouteUpdate and beforeRouteEnter to load job json from s3 before page load
- change .env files to not be git ignored
- switching configs requires updating the s3 folder in .env if we want job detail pages to work
- removes jobDetail and jobExpired
- adds a footer so we can see CLS

## Notes to Reviewer
This is only updating the job detail page.. it does not update the search page which will eventually use the same demo searchForm.


.env.production
```diff
@@ -0,0 +1,2 @@
+GRIDSOME_USE_MINIKUBE=false
```
surgiie, line 1: This repo doesnt deploy, doesnt seem like we need this file?

src/components/ApplyButton.vue
```diff
@@ -0,0 +1,79 @@
...
+        getUtmParams() {
+            try {
+                return JSON.parse(sessionStorage.getItem(UTM_KEY))
+            } catch {
+                return {}
+            }
+        },
+        getViewSource() {
+            try {
+                return {vs: sessionStorage.getItem("vs")}
```
surgiie, line 41: The `"~/services/storage"` service contains a `VS_KEY` constant you can utilize here.

src/mixins/formatLocationParam.js
```diff
@@ -0,0 +1,53 @@
+import {capitalize, words} from "lodash"
+
+export default {
```
surgiie, line 3: For the most part im okay with this mixin. 

However i have some slight concerns about opening up the door for code repetition.

Personally i think combining this mixin and and the moc one into a `~/mixins/formatRouteParams.js` seems like a good improvement cause they both duplicate the `isSlug` function.

Secondly, im not sure i like that it enforces the component utilizing it to have a `this.formData` property. IMO changing the `formatLocationParam` to accept the location as an argument would allow reuse of the function/mixin outside of the search form. So it could compute the formatted value and then it is not explicitly meant for the search form component only, it could be used for other components. For example it can be used in another component to format the location active filter/chip  on listing or job detail pages that have active filters.

src/mixins/searchForm.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export default {
+    mixins: [formatLocationParam, formatMocParam],
+    data() {
+        return {
+            formData: defaultFormData,
+        }
+    },
+    created() {
```
surgiie, line 20: The created and watch function have  the same logic, only difference is the the object being used to update the data. You can reduce the duplication here by extracting this logic to a  function that accepts the data to use to set the form data, ie:

```js
created(){
    this.initData({...this.$route.query, ...this.$route.params})
},
watch:{
    $route(newData){
        this.initData({...newData.query, ...newData.params})
    },
},
methods:{
   ...
   initData(routeData){ // or whatever you feel is name appropriate
       this.setDefaultFormData()
       this.updateFormData(routeData)
       this.formatLocationParam()
       this.formatMocParam()
  
   }
}

```

src/templates/Job.vue
```diff
@@ -1,38 +1,101 @@
...
+            meta: [
+                {rel: "preconnect", href: "https://microsites.dejobs.org/"},
+            ],
+        }
+    },
+    computed: {
+        jsonLd() {
+            return JSON.stringify(this.isExpired ? this.expiredSchema : this.schema)
+        },
+        schema() {
```
surgiie, line 65: This schema is pretty much the same on every site. We could wrap this into a exportable function if we wanted? 

So from Zarif's end: 

`getJobSchemaData(job, customData)`

This just merges this default with anything custom RR passes in?

src/mixins/searchForm.js
```diff
@@ -0,0 +1,62 @@
...
+        },
+        updateFormData(routeData) {
+            if (!isEmpty(routeData)) {
+                this.formData = this.mergeObjects(this.formData, routeData)
+            }
+        },
+        setDefaultFormData() {
+            this.formData = defaultFormData
+        },
+        mergeObjects(obj1, obj2) {
```
surgiie, line 55: This function name is confusing for what the function actually does:

image

This function overwrites key/values in object 1 that were set in object 2 key and throws away/doesnt merge keys in object 2 if they dont exist in object 1. `mergeSameKeysInto(src, from)` or something different would make it more clear what this does?

surgiie (changes requested):
Made some comments about changes, most of them can be left for another PR as they are picky/code cleanup, ive documented them in the review for future reference.

src/mixins/searchForm.js
jeremytubbs: not sure what to rename this.. ignoring for now

src/templates/Job.vue
jeremytubbs: not sure.. I am going to leave it outside of our code so it is easy to change.

surgiie (approved):

PR #163, opened 2021-10-04
[PD-7387][PD-7475] single search components

nate998877:
## What does this PR do?
These Components are direct copies from the microsites-template repo.

## Notes to Reviewer
I added no styling or anything beyond minimal functionality. Feedback on this is appreciated

## Document Changes
TODO


src/components/Form/AppMilitaryBaseSingleForm.vue
```diff
@@ -0,0 +1,47 @@
...
+        />
+        <input type="submit" value="submit">
+    </form>
+</template>
+
+<script>
+import AppAutocompleteInput from "~/components/Form/AppAutocompleteInput"
+import { MilitaryCompleteService } from "~/services/search"
+
+const Base = {base:""}
```
surgiie, line 23: shouldnt this be lowercase?

surgiie (changes requested):
Remove `Single` and `App` from component names. As we are working on cleaning up microsite-js, i think the general consensus is that we only need to prefix `App` at import time, and the file name itself doesnt have to contain `App` in its name, simplifying the file name a bit.

surgiie:
Im also a bit unsure as to why these need to exist in microsite-js. These are custom forms that should and can be implemented on Zarif's end.

surgiie:
I know theses are just demo components, but if we can avoid adding extra files to this repo, i would prefer that. That said, i think you can honestly make this generic enough to be 1 component by accepting the name of the input and query to use for autocomplete, then just bind `$attrs` to the autocomplete input to add all the input specific attributes to the input element. Something like: 

```js
<template>
    <form @submit.prevent="search">
        <AppAutocompleteInput
            v-model="input"
            v-bind="$attrs"
            :query="query"
        />
        <input type="submit" value="submit">
    </form>
</template>

<script>
import AppAutocompleteInput from "~/components/Form/AppAutocompleteInput"

export default {
    inheritAttrs: false,
    props:{
        query: Function,
    },
    components: {
        AppAutocompleteInput,
    },

    data() {
        return {
            input: ""
        }
    },
    methods: {
        search() {
            this.$router
                .push({
                    path: "/jobs",
                    query: {[this.name]: this.input},
                })
                .catch((e) => {})
        },
    }
}

</script>
```

Then you can just re use for each query type:

```js
<AppAutocompleteForm
    :query="MOCCompleteService"
    name="moc"
    placeholder="Search by moc"
    ....
>
</AppAutocomplete>

```

surgiie (approved):

PR #165, opened 2021-10-05
simplify route param formatting in searchForm mixin.

surgiie:
## What does this PR do?
The route param formatting for moc and particularly location is alot more complicated than it needs to be. The if statement checking if the params are set on `$route.params` makes it so you dont have to check if the value we are formatting is a slug or not so the `isSlug` and `!isSlug` checks are not needed because `$route.params` are always slugs, they come from the url. The location formatting also has alot if else nesting going on and doesnt need to be that complicated. That said this PR:

* Removes unneeded isSlug function and other one time methods that are only used in location param formatting.
* Simplify formatLocationParam & formatMocParam functions.
* Fixes broken import that i forgot to update in this PR https://github.com/DirectEmployers/microsite-js/pull/164

### Notes 
Moc urls should just show the moc code upper cased in the input and

the location urls should all format appropriately as before, here are some examples of all the variations you can have for url locations:

City, State:
http://localhost:8080/locations/el-segundo-ca/jobs (El Segundo, CA)
http://localhost:8080/locations/oklahoma-city-ok/jobs (Oklahoma City, OK)

Location Exact with country:
http://localhost:8080/locations/bengaluru-ind/jobs (Bengaluru, IND)

Country ISOS
http://localhost:8080/locations/ind/jobs (IND)
http://localhost:8080/locations/usa/jobs (USA)

Anything else is simply title cased:

http://localhost:8080/locations/north-carolina/jobs (North Carolina)


src/mixins/searchForm.js
```diff
@@ -97,62 +97,34 @@ export default {
...
+            this.formData.location = parts.join(" ").replace(" ,", ",")
         },
         formatMocParam() {
-            if (this.$route.params.moc) {
-                let moc = this.$route.params.moc.toUpperCase()
-
-                if (this.isSlug(moc)) {
-                    this.formData.moc = moc.split("-").shift()
-                }
+            if (!this.$route.params.moc) {
```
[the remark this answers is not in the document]
jeremytubbs: Changing the formatting does remove MOC code from other filters.. we just need to make sure that the canonical link on all those pages is correct.

src/services/routing.js
```diff
@@ -0,0 +1,41 @@
...
+
+        parts = parts.map(capitalize)
+        parts.push(last)
+
+        return parts.join(" ").replace(" ,", ",")
+    }
+}
+
+export function formatMocParam(moc) {
+    if (!moc || !isSlug(moc)) {
```
jeremytubbs, line 32: also don't need to check if `!moc`

src/services/routing.js
```diff
@@ -0,0 +1,41 @@
...
+
+function isSlug(value) {
+    return value.includes("-")
+}
+
+function isLocationAbbv(value) {
+    return value.length == 2 || value.length == 3
+}
+
+export function formatLocationParam(location) {
```
jeremytubbs, line 12: Probably should remove `Param` from both of these names now that they can be used on other inputs.

jeremytubbs (changes requested):
Minor changes.

PR #188, opened 2021-11-03
[PD-7571] Autocomplete inputs autocompletes

nate998877:
## What does this PR do?
When navigating using up/down/tab/enter the current index of the results should appear in the input & should be set as the search input when tabbing/entering. Entering worked as expected, but not up/down navigation. This PR fixes the up/down issues.

## Notes to Reviewer
To view the functionality I'm referring use the autocompleteinput @ http://localhost:8080 on qc and type in anything that returns autocomplete results eg "ind" in the location & use up/down arrows. Then checkout this branch and do the same. Notice how the display value changes where it didn't on qc.


src/components/Form/AppAutocompleteInput.vue
```diff
@@ -197,7 +200,15 @@ export default {
             }
             this.scroll()
         },
+        tabDown(){
+            this.holder={display:this.value}
+        },
         scroll() {
+            if(this.selectedIndex == this.results.length){
```
jeremytubbs, line 207: move to a separate method.. maybe named `resetUserInput` also if you just set `this.holder=this.value` you can just do once in `setValue({display: this.holder)` rather than throughout the code

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -197,7 +200,15 @@ export default {
             }
             this.scroll()
         },
+        tabDown(){
```
jeremytubbs, line 203: can be removed

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -31,6 +31,7 @@
                 @keydown.esc="blur"
                 @keydown.up="keyUp"
                 @keydown.down="keyDown"
+                @keydown.tab="tabDown"
```
jeremytubbs, line 34: can be removed

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -124,6 +125,7 @@ export default {
             result: null,
             loading: this.isLoading,
             error: null,
+            holder: ""
```
jeremytubbs, line 128: rename to `userInput`

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -145,6 +147,7 @@ export default {
             }
         }, 200),
         changeValue(value) {
+            this.holder = {display:value}
```
jeremytubbs, line 150: change to `this.userInput = value`

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -178,6 +180,7 @@ export default {
                 this.results = []
                 this.$emit("setResult", null)
             }
+            this.holder={display:this.value}
```
jeremytubbs, line 183: can be removed

jeremytubbs (changes requested):
I really like this functionality.. just some small code changes required.

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -197,7 +200,15 @@ export default {
             }
             this.scroll()
```
jeremytubbs, line 201: call `shouldResetUserInput()` before scroll

surgiie (changes requested):
Works perfect and i like these changes as well! However, im personally not a fan of the `shouldResetUserInput()` name because its naming implies it returns a boolean or should be used in a if statement. 

Consider removing the `should` word from it if were calling the function as is. `i.e this.resetUserInput()`

zarifrahat (approved):

jeremytubbs (approved):

surgiie (approved):

PR #205, opened 2021-11-29
[PD-7657] - add allow blank search prop

surgiie:
## What does this PR do?
* Add a new prop that signals provider whether or not a blank search is allowed.


src/components/Search/AppProvider.vue
```diff
@@ -58,44 +62,52 @@ export default {
     },
     methods: {
         doSearch(type) {
+            let searchData = {
+                ...this.allData,
+                ...this.immutableData,
+            }
+
+            if(this.requireKeyword && !searchData.q){
```
jeremytubbs, line 70: update this to check for any type of "query" data

surgiie: Believe the code ive implemented will now prevent a blank search if specified.

src/components/Search/AppProvider.vue
```diff
@@ -49,38 +54,56 @@ export default {
...
             }
         },
     },
     methods: {
+        shouldSearch() {
+            if (this.allowBlankSearch) {
+                return true
+            }
+            // make sure at least one input for search criteria is present.
+            let data = omit(this.searchData, ["page", "num_items", "sort", "offset", "r"])
```
jeremytubbs, line 70: this can be updated to 
```
            let data = omit(this.searchData, ["page", "num_items", "sort", "offset", "r"])
            if (! isEmpty(data)) {
                return true
            }
```

src/components/Search/AppProvider.vue
```diff
@@ -37,6 +38,10 @@ export default {
         }
     },
     props: {
+        allowBlankSearch: {
+            type: Boolean,
+            default: true,
+        },
         extraFilterData: {
```
jeremytubbs, line 45: `extraFilterData` is not used

jeremytubbs (changes requested):

jeremytubbs (approved):

PR #206, opened 2021-11-30
scale icon on zoom in military map

surgiie:
## What does this PR do?
* Add event handler for resizing icons on zoom.

## Notes 

Kudos to: https://stackoverflow.com/questions/3281524/resize-markers-depending-on-zoom-google-maps-v3

You can test with this branch:

https://github.com/DirectEmployers/microsite-template/tree/militar-zoom

http://localhost:8080/military-map


src/components/Search/Maps/AppMilitaryBaseMap.vue
```diff
@@ -278,10 +280,20 @@ export default {
...
+        zoomChanged(zoom){
+            let pixelSizeAtZoom0 = 2; //the size of the icon at zoom level 0
+            let maxPixelSize = 100;
+
+            let newSize = Math.round(pixelSizeAtZoom0*Math.pow(2,zoom));
+            if(newSize > maxPixelSize) //restrict the maximum size of the icon
+                newSize = maxPixelSize;
+
+            this.iconSize = newSize
+        },
```
nate998877, line 292: I would like to see the math here changed. 100px is a good max, but 2px is too small for a minimum. It's just black dots. 

I think 15px is a good minimum, but the specifics on that don't matter. I would just like to be able to differentiate icons even if you can't identify them. I also think the size should change linearly. The current math causes the icons to be way too large & when zooming in 1 then 1 out causes the icon to be a different size than they start with.

I've found 
pixelSizeAtZoom0 = 15;
iconSize=35;  //set in data
Math.round(pixelSizeAtZoom0+zoom*5);
gives good sizes throughout.

This isn't a perfect solution as it hard codes the icon sizes for a given zoom level & the default size won't line up if you change the default zoom level. If we care about that I think a conversation needs to be had about the full extent of what this should handle & how much we expose to RR.

surgiie: Implemented this logic.

nate998877 (approved):

src/components/Search/Maps/AppMilitaryBaseMap.vue
```diff
@@ -278,10 +280,19 @@ export default {
                 this.doSearch()
             }
         },
+        zoomChanged(zoom){
+            let pixelSizeAtZoom0 = 15; //the size of the icon at zoom level 0
```
jeremytubbs, line 284: can we make both of these props with default values?

jeremytubbs (changes requested):

jeremytubbs (approved):

PR #210, opened 2021-12-02
[PD-7670] - add new mixin to build active filters for config filter query params

surgiie:
## What does this PR do?
* Adds a mixin that can be used to build query param "active filters" when their associated url route params are not present. 

### Notes:

Associated template PR to test with: https://github.com/DirectEmployers/microsite-template/pull/13

#### What problem does this solve?

This fixes an issue that is confusing for a user experience imo:

The following is confusing from a user perspective:

Search api does not return active filters when input is given via query params. It only does so when you use filter/facet urls,

so urls like: 

localhost:8000/jobs?location=California 
localhost:8000/jobs?jobcategory=Sales

dont return active filters in the response so the front end doesnt build these as active filters/current search criteria which is very confusing.  This PR adds a mixin that will build these active filters ONLY if a url route param for the filter is not already present,
so as an example this url  `localhost:8000/locations/texas/jobs?location=California` would only ever have the active filter already returned from search api and will be ignored, but `localhost:8000/jobs?location=California` will build an active filter.

`q` is not built by search api, but i see Zarif manually add this pretty much on all sites , this mixin will also build q in addition to the config filters.


nate998877 (approved):

jeremytubbs:
closing this PR as unneeded

PR #216, opened 2021-12-07
[PD-7658] a11y updates

nate998877:
## What does this PR do?
Adds focus call for load more a11y tabbing support.

## Notes to Reviewer
This ID is hardcoded due to an inability to determine which element should be focused in the template itself. If you can find a better way to handle this it would likely be much more flexible.


src/components/Search/AppProvider.vue
```diff
@@ -137,6 +137,9 @@ export default {
         },
         loadMore() {
             if (this.response.pagination.has_more_pages) {
+                try {
```
surgiie, line 140: I personally prefer to scope this down a bit by the component's element instead of querying on that instead of the document, and ids work but i like using data attributes:

`this.$el.querySelector('[data-load-more'])`

I honestly wouldnt put a try catch on this, if we want to have a11y for our sites, id rather see it error out so Zarif can fix it.

surgiie (changes requested):
Small comment made, i need to test still.

surgiie (changes requested):
Im unsure about this code, i kinda like the idea of just emitting an event for load more instead and then letting implementer focus the button themselves:

```js

loadMore(){
    this.$emit("loadMore")
    ...
}
```

```html
<AppSearchProvider @loadMore="$refs['loadMore'].focus()" ..../>
```

PR #218, opened 2021-12-09
remove unused location file and adjust build job detail url function accordingly.

surgiie:
## What does this PR do?
* Remove duplicate unused `slugify` helper
* Remove location.js file, this is no longer used or even needed
* Remove `removeCountry` from `buildJobDetailUrl`. Data has been standardized and this is no longer needed.

### Future consideration

Move `buildJobDetailUrl`  to the `AppJobDataTransformer` component as its the only place this is being used on latest components, but we can hold off on this till all sites are updated on latest microsite-js components.  this helper was added early on to handle bad data in google talent. We have standardized our data now and we no longer need to remove country.


jeremytubbs:
Closing to wait for planning.

PR #222, opened 2021-12-28
change radius default

surgiie:
## What does this PR do?
Default radius is now 1 mile for exact location


jeremytubbs:
we need to allow Exact Location search and not always do a radius search

PR #226, opened 2022-01-10
[PD-7727] featured_jobs updates in response object

nate998877:
## What does this PR do?
Response object passed to the template should include updates to featured jobs. It currently does not. This PR fixes this.

Associated PR - https://github.com/DirectEmployers/microsite-template/pull/29


src/components/Search/AppProvider.vue
```diff
@@ -90,11 +90,18 @@ export default {
             searchService(this.searchData, this.$siteConfig)
                 .then((r) => {
                     this.updateStatus({resolved: true})
+
                     if (type === "loadMore") {
+                        if (r.data.pagination.page == 1) {
```
surgiie, line 95: Unsure of what this solves. Why would we only update featured jobs on page 1 or else concat them to the main jobs?  Featured jobs shouldnt be concatenated to the main jobs, they should be separate.

src/components/Search/AppProvider.vue
```diff
@@ -90,11 +90,18 @@ export default {
...
+                            this.response.featured_jobs = r.data.featured_jobs
+                        } else {
+                            this.response.jobs = this.response.jobs.concat(r.data.featured_jobs)
+                        }
                         this.response.jobs = this.response.jobs.concat(r.data.jobs)
                         this.response.pagination = r.data.pagination
                     } else if (type === "paginate") {
                         this.response.jobs = r.data.jobs
```
surgiie, line 103: I dont like how complicated this function is starting to get with all the if else logic. We're extracing specific properties into `this.response` from `r.data` but we honestly should just be setting the entire response so that we're not restricting anything from Zarif. Just give him control and do `this.response = r.data`

surgiie: The last else seems redundant too. Remove it.

surgiie (changes requested):

src/components/Search/AppProvider.vue
```diff
@@ -90,16 +90,10 @@ export default {
...
-                    } else if (type === "paginate") {
-                        this.response.jobs = r.data.jobs
-                        this.response.pagination = r.data.pagination
-                    } else {
-                        this.response = []
-                        this.response = r.data
-                    }
+                    let currentJobs = this.response?.jobs ?? []
+                    this.response = {...r.data}
+                    r.data.featured_jobs.map(x => (x.featuredJob = true))
```
surgiie, line 95: I prefer if this was called `isFeatured` to be consistent with its boolean type.

surgiie (changes requested):

surgiie (approved):

jeremytubbs (approved):

PR #232, opened 2022-03-03
slugged

nate998877:
## What does this PR do?
converts props from camel case to sluggified


src/components/AppCookieConsent.vue
```diff
@@ -3,8 +3,8 @@
         <transition :name="transitionName">
             <section v-if="!hasAcknowleged">
                 <slot
-                    :acceptCookieUse="acceptCookieUse"
-                    :declineCookieUse="declineCookieUse"
+                    :accept-cookie-use="acceptCookieUse"
```
surgiie, line 6: This is a slot data attribute not a prop. Undo this change.

src/components/AppDropdown.vue
```diff
@@ -45,7 +45,7 @@
                     <slot
                         :link="link"
                         :name="link.key"
-                        :isSelected="index == selectedIndex"
+                        :is-selected="index == selectedIndex"
```
surgiie, line 48: Same here, undo this change.

surgiie (changes requested):

PR #236, opened 2022-03-11
cleanup load all jobs.

surgiie:
* Cleanup loadAllJobs, current while loop nesting is making code harder to read and is gross.


src/services/search.js
```diff
@@ -136,40 +136,43 @@ export class LocationCompleteService {
 /**
  * Loads all jobs for a given origin and calls the given callback on each job.
  */
- export async function loadAllJobs(origin, callback = ()=> {}, payload = {}){
-    var error;
-    var hasMore = true
+ export async function loadAllJobs(origin, callback = null, payload = {}){
+
+    var jobs = []
     var request = {page: 1, num_items: 500, ...payload}
```
nate998877, line 142: I think num_items should default to 100. That's the maximum for GT & for a lot of sites 100 will be enough. The num_items should only be increased when it's needed to be for a large site.

nate998877 (approved):

PR #237, opened 2022-03-17
[PD-7766] smart radius search

nate998877:
## What does this PR do?
Updates the AppAutocompleteinput component to better handle showing the radius search option.

## Notes to Reviewer
shouldShowRadius covers several explicit checks for radius search including.

- Your Location
- Autocomplete option selected
- User inputs location with `city, xx` format.
Note, I did add the option for `city, XXX` for instances such as 'New York, USA'. This is controlled via `\w{2,3}` in the regex section.

The updates made to this.result and option are to allow shouldShowRadius know whether the current input is related to a dropdown selection or not. 

## Document Changes
PR - https://github.com/DirectEmployers/microsite-v2-docs/pull/11


src/components/Form/AppAutocompleteInput.vue
```diff
@@ -41,7 +41,7 @@
                 @input="changeValue($event.target.value)"
                 :aria-activedescendant="activeDescendant"
             />
-            <slot name="add-on-end" :loading="loading"></slot>
+            <slot name="add-on-end" :loading="loading" :shouldShowRadius="shouldShowRadius"></slot>
```
jeremytubbs, line 44: I don't think we should include implementation specific code in the general autocomplete input.. since we have so much extra functionality for Location Autocomplete .. I would suggest we create a new component for LocationAutocomplete that contains all the logic required for Radius select.

src/components/Form/AppAutocompleteInput.vue
```diff
@@ -220,6 +222,13 @@ export default {
         },
     },
     computed: {
+        shouldShowRadius(){
```
jeremytubbs, line 225: Make a new LocationAutocomplete component that uses the AppAutocompleteInput.. it will include `shouldShowRadius` and the computed options for radius that are in microsite-template.

jeremytubbs (changes requested):

surgiie (changes requested):
See my comments: https://github.com/DirectEmployers/microsite-template/pull/40#pullrequestreview-920675896

jeremytubbs (commented):
Sorry if I wasn't clear.. I would like us to add a LocationRadiusAutocomplete component that accepts props for the radius options not a mixin.

PR #248, opened 2022-04-06
[PD-7893] Separate data sent to search

nate998877:
## What does this PR do?
Updates the search provider to send search data as discreet query & params rather than joined into a single object. This is in preparation for a new endpoint. A separate backend PR exists to allow existing sites to function with this change. It must be merged before this can be used.

## Notes to Reviewer
To test, checkout QC on template & use this branch with `npm run symlink`. Then perform any searches that use AppProvider.
Note: the backend change allow all endpoints to accept split inputs. Since searches are the target of the new endpoint I have only updated the search provider here.
Related PR: https://github.com/DirectEmployers/microsites-search-api/pull/349

A new environment variable ENV_SHOULD_POST was added. The new endpoint's default behavior is get requests. POST requests should only be used to test config updates locally.


src/services/search.js
```diff
@@ -34,7 +34,8 @@ export function api(options = {}) {
 }
 
 function apiService(input, config, endpoint) {
-    if (isDevelopment()) {
+
+    if (isDevelopment() && process.env.ENV_SHOULD_POST === "true") {
```
surgiie, line 38: What is the purposes of the `process.env.ENV_SHOULD_POST` ? It should use POST in local dev regardless if this is set or not.

surgiie (changes requested):

PR #294, opened 2022-06-24
update map marker slot

surgiie:
### Release notes
RRR Fixed a bug in an app...


zarifrahat:
Map component was updated by Nate

PR #304, opened 2022-07-19
[PD-8002] - GA4 apply button updates

jeremytubbs:
## What does this PR do?
- Adds more info to gtag pageview
- Changes guid to be be a required property and job-data is no longer a property

## Notes to Reviewer
checkout microsite-template PR - https://github.com/DirectEmployers/microsite-template/pull/124

`npm run fresh:symlink` to test

You can visit analytics.google.com and look at the debug view to see events and page views.
image

## Document Changes
`job-data` is no longer a required prop of the ApplyButton component.. instead pass only the `job.guid` to the component.


surgiie (approved):
Not sure why my debug view looks different ( i dont see icons on the timeline like yours) but i was able to see events come through into the debug view:

image

PR #320, opened 2022-08-01
[PD-6879] autocomplete update

nate998877:
## What does this PR do?
Adds a new autocomplete service for military base searches by branch. Updates the form mixin to include a default for the militarybase option.

## Notes to Reviewer
Testing branch - https://github.com/DirectEmployers/microsite-template/tree/test-8129
Associated PR - https://github.com/DirectEmployers/microsites-search-api/pull/380


src/mixins/searchForm.js
```diff
@@ -6,8 +6,9 @@ const defaultFormData = {
     q: "",
     r: "",
     moc: "",
-    location: "",
     coords: "",
+    location: "",
+    militarybases: "",
```
jeremytubbs, line 11: I would prefer a shorter name for this query `mb` or instead of `militarybases`.. you will also need to submit the `branch` I as a select input. Setting a branch input could be used for both the autocomplete for military bases and the moc codes when implemented.

jeremytubbs (changes requested):

surgiie (approved):

src/mixins/searchForm.js
nate998877: Currently, I have the branch defined in the config field https://github.com/DirectEmployers/microsite-template/blob/717422e9cba846665d3bbf20a317fadf7600e07f/openresty-dejobs-org/components/AppSearchForm.vue#L130. I personally believe the moc autocomplete should be handled by the frontend & shouldn't affect this work, but we can have a discussion about that, if needed. Regardless of the result of that conversation, I would prefer to avoid having the input be responsible for updating multiple form bindings.

jeremytubbs:
closing because not needed.

PR #329, opened 2022-08-11
Remove alphabetical service.

surgiie:
## What does this PR do?
* Remove alphabetical service. 

## Notes
This service file only returns a function that compares object properties and doesn't actually alphabetically order an object so the nature of the service's purpose is a bit misleading. The `.sort` call [here](https://github.com/DirectEmployers/microsite-js/blob/quality-control/src/services/filterPaths.js#L36) is what actually does the sorting, thisfunction simply tells it how to sort the object. That said, the service IMO seems misleading on what it actually does and seems redundant to have as a stand alone service for a one time internal use for sorting routes.


nate998877 (approved):

src/services/filterPaths.js
```diff
@@ -33,7 +32,9 @@ function getAllFilters(filters) {
 function getCustomFilters(filters) {
     let customFilters = filters
     customFilters = customFilters.filter((filter) => !defaultFilterNames.includes(filter.name))
-    return customFilters.sort(alphabeticalOrder("display"))
+    return customFilters.sort(function (a, b) {
```
jeremytubbs, line 35: Add a comment.. one reason to extract functionality out to a named method is because it makes the code self documenting.. now I have to look at this code and think.

jeremytubbs (changes requested):

src/services/filterPaths.js
nate998877: Can't we just assign the `customFilters.sort(...)` to some variable `alphabeticalOrderFilters` & return that? Is that not self documenting enough?

jeremytubbs: yes that would be fine but a comment is also fine

jeremytubbs (approved):

PR #331, opened 2022-08-15
[PD-8152] Job Page Route Validation

nate998877:
## What does this PR do?
Adds logic to the jobMixin which updates the URL if location & title don't match the job description.

## Notes to Reviewer
This occurs after the page has already loaded & doesn't perform an actual redirect, so no 30X is sent. I'm unaware of a way to do that without disrupting the user experience.


jeremytubbs (changes requested):
This doesn't work, I would suggest moving the check to a method and putting it `jobRequest`   in the created lifecycle hook before `setJob` is called.

src/mixins/getJob.js
```diff
@@ -41,6 +42,7 @@ export default {
         if (process.client || (!process.server && this.job == null)) {
             jobRequest(this.$route.params.guid)
                 .then((data) => {
+                    this.validateUrl(data)
```
jeremytubbs, line 45:
```suggestion
                    this.validateUrl(data.data)
```

src/mixins/getJob.js
```diff
@@ -51,6 +53,17 @@ export default {
         }
     },
     methods: {
+        validateUrl(data) {
+            let job = data.data
```
jeremytubbs, line 57:
```suggestion
            
```

src/mixins/getJob.js
```diff
@@ -51,6 +53,17 @@ export default {
         }
     },
     methods: {
+        validateUrl(data) {
```
jeremytubbs, line 56:
```suggestion
        validateUrl(job) {
```

jeremytubbs (approved):

PR #336, opened 2022-08-19
[PD-8166] cookie re-engage button

nate998877:
## What does this PR do?
Cookie rework related to re-engage cookie button. Specifically, updated storage.py to use a single cookie to handle tracking of tracking cookie interaction. The value can be a string of True, False, or null (unset) which correspond to accepted, rejected, or yet to interact.


nate998877:
This work is included in 8167 & why review two prs when you could review 1

PR #338, opened 2022-08-24
[PD-8167] opt out trackers

nate998877:
## What does this PR do?
Blocked by PD-8166. This changes the default load status of trackers to enabled.

## Notes to Reviewer
Related PR -https://github.com/DirectEmployers/microsite-template/pull/161


surgiie (changes requested):
My only concern with changing the key from 2 separate "accepted" and "declined" is that this will force users who have already opted in/out to re-engage with accepting/declining cookie preferences.

nate998877:
> My only concern with changing the key from 2 separate "accepted" and "declined" is that this will force users who have already opted in/out to re-engage with accepting/declining cookie preferences.

There's the option to add some conversion code to prevent the this from happening. But is there any real impact by re-prompting & how long do we leave the conversion code?

jeremytubbs (changes requested):
This branch is appropriately based on QC. It is overwriting recent changes to the `quality-control` branch.

jeremytubbs:
> My only concern with changing the key from 2 separate "accepted" and "declined" is that this will force users who have already opted in/out to re-engage with accepting/declining cookie preferences.

@surgiie Please elaborate. There was never a need for two separate values and there has never been a way for a user to change their initial selection.

surgiie:
> > My only concern with changing the key from 2 separate "accepted" and "declined" is that this will force users who have already opted in/out to re-engage with accepting/declining cookie preferences.
> 
> @surgiie Please elaborate. There was never a need for two separate values and there has never been a way for a user to change their initial selection.


Right now, when a user engages with the cookie banner, it will set these keys: 
image


This PR, changes from this to a single key called `responded_cookie_policy` so once a site deploys these changes, a user will have to re-enage with the banner as the code for checking if a user has declined cookie preferences is no longer checking the `declined_cookie_use` key.

jeremytubbs:
@nate998877 We should keep the single key called `accepted_cookie_use` .. `responded_cookie_policy` is too generic.

src/services/storage.js
```diff
@@ -1,35 +1,39 @@
...
 }
 
 /**
- * Check if storage contains has accepted cookie use.
+ * Clear cookie in local storage.
  */
-export function acceptedCookieUse() {
-    return localStorage.getItem(ACCEPTED_COOKIES_KEY) === "true"
+export function clearCookieUse() {
+    return localStorage.removeItem(COOKIES_KEY)
```
jeremytubbs, line 28: We should not `clearCookieUse` the re-engage feature should just be component level and show the banner again allowing the user to change their choice or just show the info and link to the Privacy Policy.

jeremytubbs (changes requested):
Please change the re-engage functionality to not clear the users previous selection and only work at the component level for `showBanner`.

src/services/storage.js
```diff
@@ -1,35 +1,39 @@
...
 
 /**
  * Set cookie use in local storage.
  */
 export function setCookieUse(accepted = true) {
-    localStorage.removeItem(DECLINED_COOKIES_KEY)
-    localStorage.removeItem(ACCEPTED_COOKIES_KEY)
-    localStorage.setItem(ACCEPTED_COOKIES_KEY, accepted)
-    localStorage.setItem(DECLINED_COOKIES_KEY, !accepted)
+    clearCookieUse()
```
jeremytubbs, line 35: It is not necessary to `clearCookieUse` we can just set it.

src/components/AppCookieConsent.vue
```diff
@@ -27,37 +26,31 @@ export default {
...
-    mounted() {
-        if (process.client) {
-            this.declined = declinedCookieUse()
-            this.accepted = acceptedCookieUse()
-        }
-    },
     methods: {
         acceptCookieUse() {
-            this.accepted = true
-            this.declined = false
```
jeremytubbs, line 48: When the users is re-engaging with the cookie banner and clicking "Accept" a `window.location.reload()` is required for gtag to be enabled. Since you are using the same data prop of `acceptedCookie` to `showBanner()` I am not sure how you can tell if this is a first engagement or re-engagement. The page should not reload on first engagement but should reload on re-engagement.

jeremytubbs (changes requested):
Minor Changes. Nice work!

surgiie (approved):
Please add the reload page change that Jeremy mentioned. Aside from that, this does appear to be working as expected.

jeremytubbs (approved):

PR #342, opened 2022-08-25
Improve loadAllJobs for static build improvements

surgiie:
## What does this PR do?
* Update `loadAllJobs` to call callback after all jobs are loaded instead of after each paginated request. 
* Allow options for request payload.

### Notes

This appears to help builds alot:

image

`speedtest` has these changes and `speedtest2` is the current code.


jeremytubbs (changes requested):
Glad to see this is faster. 
Could we also just move `async function loadAllJobs` to `build.js` since it is the only place it is used? I would like to move `build.js` to a private repo some how if possible that is imported by the github action not sure if that is possible.

surgiie:
> Glad to see this is faster. Could we also just move `async function loadAllJobs` to `build.js` since it is the only place it is used? I would like to move `build.js` to a private repo that is imported by the github action... not sure if that is possible.


Made these changes, it doesnt appear to have helped much but did cut a few seconds from what  i can tell.

nate998877 (approved):

jeremytubbs (approved):

PR #346, opened 2022-09-06
[PD-8168] report cookie interact events

nate998877:
## What does this PR do?
Add Google Analytics tracking on cookie banner interact events.


src/components/AppCookieConsent.vue
```diff
@@ -43,16 +44,21 @@ export default {
                 window.location.reload()
             }
         },
+        reportInteractEvent(eventName, eventProps = {}) {
+            this.$gtag.event(snakeCase(eventName), eventProps)
```
surgiie, line 48: Is snake cased event names enforced or expected by google? Why not just pass the event names already in snake case?

surgiie (approved):

src/components/AppCookieConsent.vue
jeremytubbs, line 48: No snake case in not required.. we are using it on the Apply Link button because we are also adding an `event_label` which is title cased for display in the GA dashboard.

jeremytubbs (changes requested):
please add an `event_label` (Title cased),  an `event_category: "Cookies` and `domain: origin` to the event.

PR #351, opened 2022-09-16
[PD-8174] - auth token refactor

jeremytubbs:
## What does this PR do?
For my Auth token refactor .. I will be adding extra auth code for requests made only in development. Hopefully this refactor will make it much more clear what code is needed in development vs production.

- splits up and moves code around
- creates separate services/api entries for POST/Development vs GET/Production


nate998877 (commented):
There are client errors for autocomplete services in the AppSearchForm.

EDIT: I had the wrong template branch as there are two PD-8174 branches.

src/services/search/search.js
```diff
@@ -0,0 +1,42 @@
...
+    return search(input, `v1/solr/filters/${filter}`)
+}
+
+export function commuteSearchService(input, config) {
+    return search(input, config, "v1/google-talent/commute")
+}
+
+export function militaryMapService(input, config) {
+    return search(input, config, "v1/military")
+}
```
nate998877, line 42: These two endpoints should have config.

nate998877 (approved):

PR #387, opened 2022-11-28
[PD-7467] commute search

nate998877:
## What does this PR do?
Bugfixes & logic for commute search

## Notes to Reviewer
The object merging logic over the mergeObject function is required because it filters against a list that can't include the commute query string values.

Associated PRs - https://github.com/DirectEmployers/microsite-template/pull/226


src/mixins/searchForm.js
```diff
@@ -15,7 +15,7 @@ export default {
             formData: defaultFormData,
         }
     },
-    mounted() {
+    beforeMount() {
```
surgiie, line 18: Is there any benefit to running this in an earlier hook?

src/mixins/searchForm.js
```diff
@@ -39,14 +39,11 @@ export default {
         },
         updateFormData(routeData) {
             if (!isEmpty(routeData)) {
-                this.formData = this.mergeObjects(this.formData, routeData)
+                this.formData = {...this.formData, ...routeData}
```
surgiie, line 42: I would follow up with @jeremytubbs on this change. This function name is a bit misleading as it merges objects but only merges key/values in the 2nd object that exists as keys in the 1st object, and throws away everything else. Here's an example:


```js
mergeObj({"foo": "bar"}, {"baz": "buzz", "foo": "not-bar"})
// returns {foo: 'not-bar'}
// notice that "baz" key/value was not merged in cause it didnt exist in the first object.
```

Im under the impression @jeremytubbs made this as an intended change for preventing random url slugs/values from being merged into the payload data.

src/components/Search/AppSearchProvider.vue
```diff
@@ -47,11 +45,15 @@ export default {
             type: Boolean,
             default: true,
         },
+        isCommuteSearch: {
```
surgiie, line 48: I thought we wanted to have this be a separate provider. Im not a fan of mixing provider/search source logic like this, considering this is a google talent only thing. Open to discussion, but i would personally prefer to see this be decoupled out into a new provider then the search page should swap between providers depending on the search type.

surgiie (changes requested):

src/scripts/symlink
```diff
@@ -20,7 +20,7 @@ then
 fi
 cd "../../microsite-js"
 npm unlink @directemployers/microsite-js
-npm link @directemployers/microsite-js
+npm link
```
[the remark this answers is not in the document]
nate998877: npm link relies on the package-name defined in the package.json. Which is already `@directemployers/microsite-js`. This is correct. Defining it with `npm link @directemployers/microsite-js` causes npm to nest the package into itself, which eventually causes an OOM error.

src/mixins/searchForm.js
nate998877: I believe tt doesn't properly repopulate the form if called after it's been mounted for reasons beyond my comprehension of the vue event lifecycle. Is there a downside to doing it?

src/mixins/searchForm.js
nate998877: I understand how the method works. This led to an issue where data that is not present in the defaultFormData not being settable on the FormData on page reloads. I assumed the search API would ignore any irrelevant data sent along with the request.

src/components/Search/AppSearchProvider.vue
jeremytubbs: Yes, this should be a separate provider.

jeremytubbs (changes requested):
Please extract commute search to a separate provider.

surgiie (changes requested):

surgiie (approved):
Approving as the code itself is working and i dont have any specific changes for this feature.

jeremytubbs (approved):

PR #447, opened 2023-02-20
new build

surgiie:
- wip
- wip
- add build-job script
- wip
- wip/deployment working
- option cleanup
- uneeded var
- remove old build
- minor cleanups
- update flags
- fix asset upload
- fix gzip bug
- remove unused args

### Release notes
RRR Fixed a bug in an app...


nate998877:
>In the current build service the jobs were all pulled from solr where html_description was not available so we used the showdown dependency to convert description to html_description, since we plan to start using imports to trigger these builds, this PR removes that dependency since it is no longer needed. 

We have yet to implement this change. Do we want to wait until the imports changes to merge this?

surgiie:
> > In the current build service the jobs were all pulled from solr where html_description was not available so we used the showdown dependency to convert description to html_description, since we plan to start using imports to trigger these builds, this PR removes that dependency since it is no longer needed.
> 
> We have yet to implement this change. Do we want to wait until the imports changes to merge this?

Ive moved these changes into a separate file that way the existing code can be still used.

jeremytubbs (approved):

PR #459, opened 2023-03-21
[PD-8495] - mixin and services

jeremytubbs:
## What does this PR do?
- adds `GeocodeService` to autocomplete services
- adds `mapLocationService` to search
- adds `mapCluster` mixin for use in microsite-template

### Release notes
No breaking changes. 
related microsite-template PR - https://github.com/DirectEmployers/microsites-template/pull/308


surgiie (approved):

nate998877 (commented):
The mapCluster is missing functionality that RR needs like the ability to remove markers from the clusterer. There's also no ability to append to the cluster options without completely overriding clusterOptions

jeremytubbs:
> The mapCluster is missing functionality that RR needs like the ability to remove markers from the clusterer. There's also no ability to append to the cluster options without completely overriding clusterOptions

I don't need to remove markers in this clusterer because there are not so many. I changed AppMapClusterer in microsite-template to accept `<gmap-cluster :options="clusterOptions" :renderer="clusterRenderer()">` not sure what this can help with but didn't know I could assign the renderer to `:renderer` and it allow options to be passed another way.

nate998877 (approved):

PR #484, opened 2023-04-18
added mixin for military map

nate998877:
## What does this PR do?
Moves overrides of the mapcluster methods used by the military map into their own mixin. The overrides have to do with differences in the structure of the response such as in searchLocation. Otherwise, changes are related filtering jobs

## Notes to Reviewer
associated PR - https://github.com/DirectEmployers/microsites-template/pull/318


jeremytubbs (changes requested):
small changes

jeremytubbs (approved):

surgiie (approved):

PR #526, opened 2023-06-23
fix: only deploy sitemaps when new --deploy-sitemaps flag is present

surgiie:
### Release notes
RRR Fixed a bug in an app...


nate998877 (approved):

nate998877 (changes requested):
The default should be true

PR #566, opened 2023-08-23
[PD-8955] - add assets with regex

jeremytubbs:
## What does this PR do?

- Changes addAssets service to use regex.
- Adds new requirement - https://www.npmjs.com/package/he for HTML entities.


jeremytubbs:
> When testing on the template I noticed http://localhost:8080/west-richland-wa/family-medicine-physician/840C6BF9F5E2488D94AEBB3E9A026B2A/job/ is failing to add assets.

I believe this is a different use of "assets" that is only a chunk of html to be used and not a text replacement.

jeremytubbs:
Closing for now because this doesn't work with static builds. `Error:  DOMParser is not defined`

nate998877 (approved):

PR #575, opened 2023-09-05
[PD-8997] UTM Parameter fix for AppApplyButton

nate998877:
## What does this PR do?
Fixes an issue where UTM parameters aren't properly applied to Apply urls. 

On first page load the `utm_` and `vs` where not present in the URL.

## Notes to Reviewer
You can use https://test-pd-8997--openresty-dejobs-org.microsites.recruitrooster.com/jobs to test


src/components/Jobs/AppApplyButton.vue
```diff
@@ -49,8 +58,21 @@ export default {
                 return {}
             }
         },
+        setApplyUrl() {
+            if (this.mailto) {
+                return this.mailto
```
jeremytubbs, line 63: need to set `this.applyUrl = this.mailto` and return

src/components/Jobs/AppApplyButton.vue
```diff
@@ -70,26 +92,11 @@ export default {
         },
         openApplyUrl() {
             if (process.client) {
+                this.setApplyUrl()
```
jeremytubbs, line 95: remove extra call to `setApplyUrl`

jeremytubbs (changes requested):
fix tests with  `async () => {` and ` await wrapper.find('a').trigger('mouseenter')`

surgiie (approved):
Nothing in particular stands out to me that Jeremy hasnt already pointed out. Please address Jeremy's feedback.

src/components/Jobs/AppApplyButton.vue
nate998877: The extra setApplyUrl is required as the initial hover won't always set the correct url depending on user behavior (loading page hovering on apply button & clicking, hovering quickly before sessionstorage is set). This appears to always set the correct url.

jeremytubbs: correct needed for firefox browser

jeremytubbs (approved):

PR #585, opened 2023-09-18
[PD-9018] duplicate requests

nate998877:
## What does this PR do?
The v2.0.0 change can be removed. While debugging, NPM crashed several times when attempting to parse "2.0.0d2". left in w/ the idea we can just not iterate the version until we're ready for a release.

I believe the issue was caused by this route watching code persisting across route changes, but unmounting then remounting the provider itself. This resulted in mounted & the watcher triggering doSearch. 

Can be tested using https://test-pd-9018--openresty-dejobs-org.microsites.recruitrooster.com/partner/providence/profession/physiciansurgeon/jobs/


surgiie (changes requested):
This does appear to fix the duplicate request but isnt this the search that is done when using the "Load More" pagination?

jeremytubbs (approved):
not sure why this works but "load more" and "simple" pagination both appear to still work.

surgiie (approved):

